PR #19, opened 2018-05-31
4. add the second Flecsale example

JieyangChen7:
Previous Flecsale can only use one node (process). So, we added a second Flecsale example that can utilize multiple nodes (processes).


trandles-lanl (approved):
Approve.  We'll test this AWS example in the future.

paulbry (approved):
New example file added, no core changes.

PR #20, opened 2018-05-31
5. add Dockerfiles used to build examples

JieyangChen7:
I added Dockerfiles that were used to build examples (Flecsale, Flecsale2, VPIC, and ParaView-Catalyst).


trandles-lanl (approved):
Approved as is.  We'll replace these examples relying on 3rd party repositories with BEE-curated repositories in the future.

paulbry (approved):
See earlier comments for additional information.

PR #21, opened 2018-05-31
6. Add BeeSwarm code

JieyangChen7:
This pull request contains the code changes that enables CI scalability test (BeeSwarm). 
There is a new installer dedicate for CI. Although currently most of the scripts in it are identical to the regular installer. I create a separate installer just in case we want to add more scripts specifically for CI in the future. Other changes include modification to the bee-OpenStack launcher to add a scalability test mode.


paulbry (approved):
From reading the paper and running a few limited tests using BeeSwarm I feel we should merge this pull. There is no adverse affects on the main branch and merging this request will allow for more use of BeeSwarm.
As we use it more, if there any enhancements identified then those could be under taken at that point.

pagrubel (approved):
I've looked through the changes, they look fine. I don't have a way to check them but they don't affect the rest of bee so lets merge this.

PR #22, opened 2018-05-31
2. Move bee_conf.json file

JieyangChen7:
The bee_conf.json is used to store the current port number that let bee_launcher and bee_orc_ctl talk to each other. The port number can change every time the bee_orc_ctl is launched. Originally this file is placed as same as bee_launcher.py. However, this creates a potential problem when we push code changes to repo. So, we choose to move this file to ~/.bee. The file is created during installation time.


paulbry (approved):
Ignore my comment from yesterday, that was a mistake in my PATH. Successfully ran Charliecloud examples for lammps__cc_gen, lammps_cc_mpi, and flecsale_gen (mpi ignored for due to other questions).

trandles-lanl (approved):
Conflict addressed.

PR #33, opened 2018-06-19
fix outdated docker installation script

JieyangChen7:
This pull request fixes the outdated installation script in the bee-vm/aws image builder.


paulbry (approved):
Approved, this is the supported way.
I added a note so it's clear that it only support 14.04 so we don't try to use it on a different version / distribution.

pagrubel (approved):

PR #35, opened 2018-06-22
Fix potential problem on multi-node run

JieyangChen7:
(1) This PR contains a bug fix that may prevent MPI run correctly on a multi-process/node environment. This bug may affect BEE-VM/AWS/OpenStack launcher. Since we run everything as root inside the container, the ssh keys need to be copied from the default user in the container (e.g. beeuser), otherwise, we may encounter an MPI error.
(2) I also add a performance optimization for BEE-OpenStack Launcher that speeds up its launching speed by 30% - 50%. It optimization basically allows worker instances to launch at the same time as the master instance.


paulbry (approved):
My tests related to VM worked, I cannot properly test the AWS functionality.

pagrubel (approved):

Issue #52, opened 2018-08-24
Encoding causes key error for unavailable task configurations


paulbry:
If you leave out the `mpi_run` or `general_run`  in the beefile it will lead to a `KeyError` from at run time.

To fully account for these options being left off there is minor work that would need done to each BEE module; however, I'm only going to aim for including this in the upcoming Bee-Charliecloud bug fixes.

Though I will need to make changes to both the launcher/composer. These changes will not adversely affect the other Bee implementations (OS, VM). Instead now if the `mpi_run` key is not available it will be ignored.
If desired I could highlight a warning message (in yellow) that the task configuration was not found if there is the requirement.

Not really a big change but since I'm making minor corrections including this was relatively easy

Issue #51, opened 2018-08-24
Move and rename `delete_after_exec`


paulbry:
The `delete_after_exec` function was added to BEE-Charliecloud to allow for the user to remove the container (simply via `rm -rf ... `) upon the completion of the workflow.

It makes sense to move these under the `container_conf` both because it logically makes more sense when building the file as well as the fact this functionality is only available for BEE-CC. So a `.beefile` would look like:
```JSON
"container_conf": {
    "container_path": "/path/to/container.tar.gz",
    "remove_after_exec": true
}
```
I have also change the name from `delete_after_exec` to `remove_after_exec` so the intended functionality would be differentiated from any of BEE's concept of deleting a task.

@pagrubel Does the change is name also work for you?

Issue #50, opened 2018-08-23
Enable flags for mpi_run like was done for srun_run in Bee charlicloud launcher.


pagrubel:
To run with mpirun on systems that may not have slurm this would be required to run using mpi. For now the flecsale example using mpi_run runs on 5 cores on two nodes and it should be splitting 5 over the 2 nodes. We need to add the flag that makes it possible but that is not available for mpi_run. This gives the user more flexibility.

Issue #49, opened 2018-08-23
Node_List cannot be left blank when using BEE-Charliecloud


paulbry:
Leaving Node_List blank will cause unexpected errors when using BEE-Charliecloud.
When flow is involved this is expected as it is required to keep from running on unintended nodes but for a single task it would probably make sense to correct this and the entire allocation should be used

I aim to correct this as part of `ipt_bug_fix` branch, let me know if you think functionality should be different.

PR #53, opened 2018-08-31
Ipt bug fixes

paulbry:
These are changes focused on both bugs that have been identified while using the recent BEE-Charliecloud update (#52, #49 , #51, #24) as well as some minor enhancements (#50). Almost all changes are focused on the BEE-Charliecloud launcher and should not have any negative affect on other launchers. There were minor alterations to high level modules (`bee_launcher.py` and `bee_composer.py`) but testing I have done so far has shown these do not affect other launcher. Please focus testing on Charliecloud and at least taking a passing glance to ensure it will not affect other aspects of BEE.

Change log:
- `mpi_run` and `general_run` now both have the ability to utilize the `flags` key in the same fashion as `srun_run`. This allows the user to define custom key/value pairs that will be used when invoking the defined script. See the BEE-Charliecloud documentation for an example, also you should note that at this time it is only supported by `bee_charliecloud_launcher.py`
- Minor modification when using Pyro4 to use an environmental variable that BEE is responsible for setting as opposed to the `-k` flag. This was being deprecated and we would receive warning when using newer versions of the module.
- When writing the `.beefile` keys such as `mpirun`, `batch_mode`, `general_run`, and `terminate_after_exec` will no longer be required.
- The `container_tar2dir` key can not be used to define the directory used when extracting the Charliecloud container. By default it is `/var/tmp` so you may not have a reason to use it.
- `node_list` (found under `exec_env_conf { charliecloud_conf ...` is no longer required. However, leaving it blank can lead to unintended & negative performance when utilizing BeeFlow.
- Minor changes to the existing BEE-Charliecloud examples to highlight some of the changes in this PR.
- Updates to BEE-Charliecloud main/example `README.md` files to reflect all changes.

I want to plan a couple more tests using the Blast example and possible update the associated `.beefile`; however, what testing I completed this morning looked promising enough to release the PR. Please let me know if you run into any problems.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ __pycache__
 ./bee-image-builder/bee-vm-and-aws-image-builder/ssh_key/config
 ./bee-image-builder/bee-vm-and-aws-image-builder/ssh_key/id_rsa
 ./bee-image-builder/bee-vm-and-aws-image-builder/ssh_key/id_rsa.pub
+.idea
+.idea/vcs.xml
--- a/bee-launcher/bee_charliecloud_launcher.py
+++ b/bee-launcher/bee_charliecloud_launcher.py
@@ -111,8 +111,9 @@ class BeeCharliecloudLauncher(BeeTask):
                 bee_cc.master = self.__bee_cc_list[0]
 
             # Remove erroneous comma (less impact), for readability
-            if self.__hosts_mpi[-1] == ",":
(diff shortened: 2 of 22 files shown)


pagrubel (approved):
Tested charliecloud changes, works on fog and darwin. The mpirun problem is with the setup on woodchuck and fog. Use srun when slurm is available.

Issue #56, opened 2018-10-25
BEE-Charliecloud untarring between tasks


pagrubel:
If a workflow needs to reuse a container there is no way right now. BEE untars the image each time. We need a mechanism to allow re-using the image.

PR #59, opened 2018-11-07
Use existing image solves issue #56

pagrubel:
This pull request gives the ability to use already untarred images for Bee-Charliecloud workflows. See the example in examples/bee-composer-example/blast-repeat/bee-cc/ and the README file.


paulbry (approved):
I didn't initially read the README.md ... so the repeated usage of the container only appears during specific phases. Please ignore my earlier comment regarding 'remove_after_exec'

The added key/value works for bee-charliecloud

JieyangChen7 (approved):
I approve this pull request. I believe the extra key in the beefile should work as expected based on my code review. However, I wasn't able to actually run it on multiple nodes on my local cluster machine since our local machine does not have user namespace support. I already requested for enabling user namespace but it might take a while for the system admin to update the settings of the system.

PR #62, opened 2018-12-28
Installation script modified to work with virtualenv

### LINKED ISSUE #61: Create BEE installation script that will work with virtualenv
The default install.sh script uses pip with --user to install thing in the user's ~/.local directory.  Create a modified version of install.sh to support users who prefer to use virtualenv.

trandles-lanl:
Fixes #61

### DIFF
--- a/install-virtualenv.sh
+++ b/install-virtualenv.sh
@@ -1,0 +1,38 @@
+#!/bin/sh
+
+# Install necessary libraries
+pip install pyro4
+pip install boto3
+pip install termcolor
+pip install tabulate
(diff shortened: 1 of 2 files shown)


pagrubel (approved):
I approve this pull request. Let's merge it.

paulbry (approved):
I should have thought more about this originally but it's only good temporarily. I think this method will help with the demonstration (hopefully won't reboot) but long term is not ideal.
Still if you want to incorporate these changes along with the support for virtualenv this should be okay.

PR #69, opened 2019-06-12
Deleted old content. Created new directories. Added README files so d…

trandles-lanl:
…irectories are not empty.


pagrubel (approved):
Fine with me. Its as we discussed in the meeting. I say go ahead and merge.

PR #80, opened 2019-06-26
Pipenv setup

### LINKED ISSUE #79: Set up developer linting environment
We need to make it easy to lint our Python code while writing it. Our linter options include `pylint`, `pep8`, `pycodestyle`, and `pyflakes`; we can use `pydocstyle` for docstrings. It's possible to use multiple of these together using `pylama` and these can all be installed using `pip`.

Whatever we choose to use, we need to make it easy to install and get started using these tools. We should create a shared config file to configure what exceptions we choose to make to PEP 8 and related standards.

We should also make it easy to install the tools that we need without too much manual labor. We can achieve this through any of the following means, in order of preference:
* Use `pipenv` to create a virtual environment and install all the needed packages
    * Can be installed with Homebrew
    * Simple commands; easy to create and activate virtual environments
    * Can separate the developer and user environments
    * Can share a `Pipfile.lock` to synchronize our packages/dependencies
* Create a script to automate the installation of packages
    * More lightweight
    * Harder to maintain
    * Does not support virtual environments as naturally as `pipenv`
* Manually install the required packages
    * No

Boogie3D:
Add Pipfile, documentation, and pylama config to master. This PR resolves #79.


mcpherson (approved):
This is good. We may need to fine tune the instructions to describe how `pipenv` builds the environment from Python versions that are _already available_ on your system. That just means that you must have Python 3.7+ installed. `pipenv` will not go download it for you.

If we want to develop/test with multiple versions of Python, we should use Docker. That would likely be compatible with our CI system

PR #83, opened 2019-07-08
Sample Neo4j databases

mcpherson:
This branch adds code (and documentation) to produce and use two simple Neo4j databases for testing. There are two databases:

- `blast.neo4j`: represents our simple BLAST workflow
- `hpc.neo4j`: represents a parameter study with an analysis step at the end

The `README.md` file in the `test/workflows/neo4j` directory should have enough information so that a developer can duplicate their creation, export, and import.


Boogie3D (approved):
The `.neo4j` files are probably fine to have on here. Don't know if architecture will cause issues later but we can just regenerate them locally if it does.

pagrubel (changes requested):
Al's agreed to make some changes to this, to make it a little clearer. In the meantime its available in the branch.

PR #85, opened 2019-07-19
Replace Pipenv with Poetry

### LINKED ISSUE #84: Replace Pipenv with Poetry
Pipenv has limited functionalities, such as a lack of actually packaging capabilities and for doing relational comparisons of required Python versions other than checking equality (i.e., `pipenv check` fails if someone is using Python 3.7).

Since we can test different Python versions with CI, requiring any and all contributors to use Python 3.6.8 is unnecessarily restrictive and can hamper development progress in the future. I propose that we replace our Pipenv dependency management system with Poetry--an alternative that uses the more flexible and [standardized](https://www.python.org/dev/peps/pep-0518/) `pyproject.toml` configuration file to replace both `Pipfile` and `setup.py`.

Like Pipenv, `poetry` can create a virtual environment based off the system Python install or that specified by `.python-version` (used by Pyenv). It supports many of the same commands as Pipenv (and more) and seems likely to be a more successful project in the long-run (some Pipenv [developers'](https://www.reddit.com/r/Python/comments/bklroc/why_im_not_collaborating_with_kenneth_reitz/) [attitudes](https://github.com/pypa/pipenv/issues/1050#issuecomment-346203646) are reminiscent of the ill-fated [Antergos](https://forum.antergos.com/topic/10895/antergos-not-installing-take-a-deep-breathe) and [aurman](https://www.reddit.com/r/archlinux/comments/9aotjr/aurman_no_longer_being_maintained_publicly/) projects).

If we decide to go this route, I can spend the time to modify the repo and to get everyone's environments set up again without Pipenv.

Boogie3D:
To implement the replacement of Pipenv with Poetry:
* `setup.py` and `Pipfile` are replaced with `pyproject.toml`
* `doc/pipfile.md` is rewritten as `doc/poetry.md`
* `poetry.lock` is committed to synchronize dependency versions
* `Pipfile` removed from `.gitignore`
* `*.egg-info` added to `.gitignore`

This PR resolves #84.


mcpherson (approved):
Merge it.

pagrubel (approved):
Merging this!

PR #102, opened 2019-12-02
Temporary hacks to workflow interface API

mcpherson:
These changes allow the current workflow interface API to be used for development of our servers. We anticipate a refactor of the workflow interface within a month.

I did not write additional tests for these changes (since they're temporary hacks) but @Boogie3D tests still run successfully though there are a few glitches that might have to do with versioning of Python and the test framework. We'll resolve those when we work on the refactor.


pagrubel (approved):
I'm fine with these changes for an intermediate fix, please make an issue to fix the hack for setting the task id.

PR #108, opened 2020-01-08
Worker interface pat

pagrubel:
Worker interface with an abstract class for work load managers, with default as Slurm including unit tests for the three initial methods, submit_job, query_job & cancel_job.
Issues for work to be added shortly after this pull request is merged:
      - Adding a common data structure for output from the methods (suggested: a named tuple).
      - Using config for the work load manager (enabling easy method to other than Slurm)


Boogie3D (approved):
Looks good to me.

PR #111, opened 2020-01-15
Yet Another Workflow Interface Refactor

### LINKED ISSUE #101: Redesign workflow interface API
The more I try and fix the current workflow interface, the more it seems that we should redesign it. The major problem is a duplication of state between the database and in-memory (Python) data structures. Here's (roughly) how the current interface, `WorkflowInterface`, operates:

There are two major Python objects, `Task` and `Workflow`. Basically (forgetting for the moment requirements and hints) we create a bunch of `Task`s (`create_task`) and these are used to create the `Workflow` (`create_workflow`). The `create_workflow` method assigns IDs to the tasks and determines dependencies (the key aspect of the workflow). It determines dependencies based on shared inputs and outputs (per CWL specification). Task dependencies are specified as a list of task IDs (`bee_init` and `bee_exit` are created and `bee_init`'s dependencies set as well). Now we have an in-memory representation of the workflow.

We then use the `load_workflow` method to insert the workflow into the database. It uses the previously created `Workflow` object to do this. After the database workflow is created, we use `initialize_workflow` to set `bee_init`'s state to `READY`.

Now we can _query_ the loaded workflow (database) to execute the workflow. This is _currently_ done by calling `get_workflow` which uses the database to re-create a (new) `Workflow` object that represents the state of the workflow at that point. Once we have that, we can look for tasks and get their dependencies (`get_dependent_tasks`). The duplication problem should be apparent. We use the in-memory `Workflow` object to find a task, but then query the database to find its dependencies! This quickly gets confusing. Also, we need to call `get_workflow` every time we suspect a change in the database (since we're polling and not using callbacks). This results in the creation of a new `Workflow` object (of the entire workflow).

My current thinking is that we should separate the creation of a workflow from its operational usage. We can still create a workflow as currently done, but once it's in the database we use an atomic query-based API instead of copying out the contents of the database to Python representation of the entire workflow. We will likely still need `Task` objects, but maybe we should have two kinds: `CwlTask` (for building the workflow) and `WfTask` for tasks in the database. Note that `WfTask` is a _read only_ object. Once the workflow is built (inserted into the database) we never _write_ or _update_ the database by passing in a modified task object. Instead, we use an interface to set modified properties (e.g. `set_task_complete` to set its state to `DONE`).

I haven't got this entirely figured out, and I'd like to work with @Boogie3D on the re-design, but here's how it might work:

The parser uses some _workflow creation_ API to build a representation of a workflow and insert it into the database. This could use most of the code and APIs that already exists. Once that's done, and the workflow is in the database, the database represents the __truth__ as we operate on the workflow.

The orchestrator will then use a query-based API to operate on the workflow:

```python
wfi = get_loaded_workflow()          # "pointer" to database workflow
head = wfi.get_head_task()           # `bee_init`
tasks = head.get_dependent_tasks()   # these tasks are "pointers" to database tasks
for t in tasks:
    wfi.exec_task(t)                 # send tasks off to task manager
# async loop loop to check for completed tasks
# get dependencies of completed tasks
# rinse and repeat
```

Much more thinking to be done.

Boogie3D:
This resolves #101. See details [here](https://github.com/lanl/BEE_Private/issues/101#issuecomment-573235274).


mcpherson (approved):
I have tested this with my modified `by_hand.py` example and all APIs that I used worked OK.

PR #143, opened 2020-04-15
Fix slurm unit tests:

pagrubel:
- Slurm worker now uses DockerRequirement for the charliecloud container.
- The task object passed to the slurm worker.
- Slurm unit tests do not use charliecloud at this time.


trandles-lanl (approved):
It all looks good to me. Good job. :)

rstyd (approved):

Issue #161, opened 2020-06-04
Task Manager should use Slurm REST API instead of PySlurm 


rstyd:
PySlurm is no longer working on fog and will soon no longer work on any of our other systems. We should switch to using the REST API instead.

PR #171, opened 2020-06-18
Slurm worker rest

rstyd:
Addresses #161


trandles-lanl (approved):
Tested on fog. It works as expected.

pagrubel (approved):
Will add an issue to fix unit tests for slurm worker

PR #185, opened 2020-06-29
Modular container runtime 157

pagrubel:
This is the initial Container Runtime Abstraction class, there are still some configuraion requirements. Enables to use other Container Runtimes.


trandles-lanl (approved):
Runs the ```cf.cwl``` workflow to successful completion.

PR #195, opened 2020-07-08
Add logging capabilities to BEE components

rstyd:
This PR addresses #193. It adds logging to all the various BEE components.

### DIFF
--- a/beeflow/common/config/config_driver.py
+++ b/beeflow/common/config/config_driver.py
@@ -159,7 +159,8 @@ class BeeConfig:
         :type relative_path: string, path to file
         """
         # Discard redundant paths, iff Windows, replace(\,/)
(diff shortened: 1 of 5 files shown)


pagrubel (commented):
@rstyd Is there a way to get the stdout of the servers, i.e. tm and wfm? 
Should slurmrestd.log be in the directory with the other logs? by default?
I had to checkout pyproject.toml from the develop branch. Please make sure that stays fixed.

qwofford (approved):
Logging functionality worked for me. Didn't break anything.

pagrubel (approved):
Logging works for me. Only logs the Flask output. I'd like to see logs for stdout of the tm and wfm. I guess we could have an issue for that though.

PR #223, opened 2020-09-21
 Handle exceptions in Task Manager for slurm worker.

pagrubel:
* Improve passing exceptions and failures to the task manager for task actions.
     Only returning pertinent data for submit, query and cancel, no success code.

   * Remove unused cancel_job function.

   * Rename check_task to process_queues since it processes whatever is in the job queues.

### DIFF
--- a/beeflow/client/gc.cwl
+++ b/beeflow/client/gc.cwl
@@ -51,7 +51,7 @@ steps:
       outputs:
         outfile: stdout
       stdout: counts.txt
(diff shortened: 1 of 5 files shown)


qwofford (commented):
Something may be off here? I'm not sure, but I don't see counts.txt in my outputs.

qwofford (approved):
Parser changes are affecting the gc.cwl, but that's already in develop so a bugfix to come from parser.

PR #225, opened 2020-09-23
Scheduler design

jtronge:
I added tensorflow as a dependency so it will need a poetry update.

If you'd like to test it, you'll need a trained model. I uploaded a test one [here](https://drive.google.com/file/d/1U6Zvy4Sh1fTPE11JabyJJqr6ezEEhuO7/view?usp=sharing). It should be untared into the root of the repo, after which tests can then be run with `pytest test/test_scheduler*.py`.


rstyd (approved):
Looks good to me!

qwofford (changes requested):
Overall, all this scheduler work is super-clean. There are some issues with BEEStart that prevent it from launching the scheduler on the fog front end. I will work on that and push to this branch when it's fixed. :+1:

qwofford (approved):

PR #233, opened 2020-10-07
Workflow manager cleanup

rstyd:
This PR cleans up some error messages, fixes some small output stuff, cleans extraneous files in the WFM directory, and makes the server->wfm change we discussed some time ago.


pagrubel (approved):
Fixes looked good. 
Added some pylama fixes and tested on case and summit (and fog where some issues with the newer slurm exist on the cancel, the behavior is the same as develop there)

Issue #232, opened 2020-10-02
WFM -- Scheduler REST Interface


jtronge:
I didn't see another issue for this so I thought I would make one now that the initial scheduler code has been merged into develop. We need to add the REST api connections to the scheduler in the WFM and then make changes to the scheduler as needed. We will probably also need to make additions/updates to the WFM -- Task Manager interface in order to pass along scheduling info, but that may be better as a separate issue.

PR #237, opened 2020-10-13
WFM Scheduler Integration fully functional

rstyd:
This PR addresses #232. It connects the workflow manager with the scheduler. The WFM doesn't actually _do_ anything with the information at the moment just because we require a more complicated workflow.


jtronge (approved):

pagrubel (approved):
I fixed the order of process startup in BEEStart and the clamr example works well. I've gone through this and all seems fine. I'd like to be able to have an option that skips the scheduler especially for small workflows that one knows will run on one cluster.

PR #239, opened 2020-10-15
Workload scheduler/issue170

### LINKED ISSUE #170: Develop an Abstract Class for using Work Load managers such as Slurm
Develop an Abstract Class for the HPC Workload system on various clusters, such as Slurm, LSF etc.
- [x] Include a configuration setting in the task_manager section called [workload_scheduler] with default Slurm
- [x] decide if we want to continue to use a template file, this may not work well with our scheduler 
          https://github.com/lanl/BEE_Private/issues/170#issuecomment-681108876 
- [x] add Worker for LSF

pagrubel:
Addresses issue #170 has an abstract case for workload_scheduler and implements driver for LSF. Workload_scheduler is specified in DEFAULT section of user configuration file.
Adds a job_template path to user configuration if needed. 
BeeStart only starts slurmrestd if Slurm is the workload scheduler.

Directions to run on summit in doc.


trandles-lanl (approved):
Tested on Summit today. Works fine.

qwofford (approved):
Tested on Fog, works as expected there. Not testing on Summit due to Poetry environment issues. Opened #240 to address the larger issue, we need a consistent environment for BEE components, and we could solve that once per BEE version rather than once per BEE developer per version.

PR #263, opened 2020-11-10
Fix delete bug

rstyd:
This fixes an error in the task manager that fails to delete all tasks in the queue if there's more than 2.


pagrubel (approved):
Tested and approved.

trandles-lanl (approved):
I don't have a workflow handy with more than 2 steps to test this. I did run the CLAMR-ffmpeg workflow and cancel worked. As far as I can tell nothing broke. I'll approve based on @pagrubel's approval and my own limited test.

PR #264, opened 2020-11-10
Tests + configuration bugs worked out

rstyd:
This PR adds a test suite for the WFM and TM.


trandles-lanl (changes requested):
My comments in task_manager also apply to wfm

pagrubel (changes requested):
To get the tests to work I moved the test directory into the beeflow directory and imported from beeflow.test.mocks

trandles-lanl (approved):
This looks good to me now.

One suggestion: Please add a couple lines to the test/readme.md file explaining how to run pytest against the new tests. An actual example command line would be nice.

pagrubel (approved):

Issue #229, opened 2020-09-29
Refactor Configuration Code


rstyd:
The configuration code currently stands at around 43 lines of code and adds significant bloat to components in which it is included. We should refactor it. I'm thinking some sort of function call.

PR #273, opened 2021-01-12
Add default ports and offset to configuration driver

pagrubel:
This PR satisfies #229 
Please look it over. I would like to change the scheduler also, however it starts up differently than the other components.
I believe it is ready to merge, but am adding the WIP label in case we want to include the changes in the scheduler.


qwofford (approved):
Nice! I really like how this PR centralizes all the default ports in config_driver.

I have not tested, but assuming someone else has...this all looks good to me.

rstyd (approved):

PR #283, opened 2021-01-26
restoring build files from commit c7b2e1b

qwofford:
Turns out I was responsible for this mishap.

When I copied files into their new locations, I simply missed the `build_interfaces.py`.


pagrubel (approved):
I reviewed these changes and tested on fog using slurm and charliecloud. Everything seems good.

Issue #151, opened 2020-04-29
Run TM and WFM with WSGI


rstyd:
We need to figure out how to run the task manager and workflow manager with WSGI using something uWSGI or nxginx.

PR #284, opened 2021-01-28
Add UWSGI Support

rstyd:
This PR addresses #151. Currently, there's an issue where cli.py starts the container more than once, and this causes the component sockets to be unavailable.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,6 +60,7 @@ jsonpickle = "^1.3"
 requests-unixsocket = "^0.2.0"
 tensorflow = "^2.3.0"
 matplotlib = "^3.3.1"
+uwsgi = "^2.0.19"
 
 [tool.poetry.dev-dependencies]
 # Developer dependencies
--- a/src/beeflow/cli.py
+++ b/src/beeflow/cli.py
@@ -14,6 +14,7 @@ import getpass
 import os
 import shutil
 import subprocess
+import string
 import sys
 import tempfile
 import time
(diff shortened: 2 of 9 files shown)


qwofford (commented):
Looks good! It does seem like there might be some config template outputs that don't need to be in the repo, only a concern because there are hard-coded ports associated with @rstyd.

pagrubel (commented):
beeflow --config-only no longer works

beeflow --config-only
Traceback (most recent call last):
  File "/yellow/users/pagrubel/BEE/BEE_Private/src/beeflow/cli.py", line 298, in StartNginx
    bc.userconfig['nginx']
  File "/usr/lib64/python3.6/configparser.py", line 959, in __getitem__
    raise KeyError(key)
KeyError: 'nginx'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/yellow/users/pagrubel/.cache/pypoetry/virtualenvs/beeflow-ApM-ipUD-py3.6/bin/beeflow", line 33, in <module>
    sys.exit(load_entry_point('beeflow', 'console_scripts', 'beeflow')())
  File "/yellow/users/pagrubel/BEE/BEE_Private/src/beeflow/cli.py", line 479, in main
    proc = StartNginx(bc, args)
  File "/yellow/users/pagrubel/BEE/BEE_Private/src/beeflow/cli.py", line 305, in StartNginx
    bc.modify_section('user','graphdb',graphdb_dict)
NameError: name 'graphdb_dict' is not defined

pagrubel (changes requested):

pagrubel:
Should this PR be closed? @rstyd

Issue #285, opened 2021-01-28
bugfix: import path inconsistencies


qwofford:
I found another case of import inconsistencies in `parse_cwl.py`.

I'm creating this issue to track a sweeping fix for this. The steps to fix this across the board are as follows:

1. Create a manifest of all files and directories that exist below `src/beeflow` in the packaging structure.
2. Grep through all source files for instances of `import x` or `from x`, where `x` is a file or directory from step 1
3. String replace `import x` with `import beeflow.x`
4. String replace `from x` with `from beeflow.x`

PR #286, opened 2021-01-28
make import paths relative to beeflow.

qwofford:
Found more instances of inconsistent paths observed by @pagrubel. Made bugfix for sweeping change to all affected files.

See #285 for details on how this was done.


pagrubel (approved):
All paths are consistent now and tests run including pytests using lsf_worker that failed when running on summit prior to this fix.

Issue #275, opened 2021-01-14
cli: Remove hardcoded subprocess runs to launch beeflow from any location.


qwofford:
In order for the cli to work as a package, we need to remove our hardcoded subprocess.run launches:

example:
https://github.com/lanl/BEE_Private/blob/ef1d065d5b657204dd7e948a367b07fbb742776b/src/beeflow/cli.py#L233-L235

This refers to a path relative to the CWD, so in this case, if you aren't in `src/`, the subprocess will fail.

PR #288, opened 2021-02-02
Fix paths in cli.py and make Tensorflow optional

jtronge:
I've modified the cli.py script to launch the Task Manager, Workflow Manager and Scheduler using the module option. This addresses [Issue #275](https://github.com/lanl/BEE_Private/issues/275) and should allow us to launch BEE from any location.

Also, I went ahead and made Tensorflow and matplotlib optional. You should be able to install them by running `poetry install -E scheduler_extras`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,8 +58,13 @@ pylint = { version = ">=2.3.1" }
 APScheduler = "^3.6.3"
 jsonpickle = "^1.3"
 requests-unixsocket = "^0.2.0"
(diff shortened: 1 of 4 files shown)


qwofford (approved):
Tested and working. I like this PR a lot!

pagrubel (approved):
I did test this on summit as well. Nice PR. We might want to fix src/beeflow/tests/test_wf_manager.py to work when tested from the main repo directory.

PR #295, opened 2021-03-02
Change prints to log messages for task manager and workers.

### LINKED ISSUE #298: logging: Simplify log creation parameters
Currently, the logging tool expects an entire BeeConfig object as a parameter, but then the only thing in needs from that BeeConfig object is the bee_workdir, which establishes the base path for the log file mount point.

No reason to pass a complex object, jsut need bee_workdir.

https://github.com/lanl/BEE_Private/blob/ffc74ce366fa098f054952ad6ddcf9d61c6a59bf/src/beeflow/common/log.py#L129-L153

pagrubel:


qwofford (approved):
Tested working for me ( after merging #299)

PR #299, opened 2021-03-04
Resolves issue 298

qwofford:
I tested the SlurmWorker logging and it looks right to me.

One thing I'm doing with logging filename convention, is using snake_case for bee component log files, and CamelCase for object log files. We can re-think that, but if the log file names look inconsistent that's why.


pagrubel (approved):
Looks good, thanks!

PR #305, opened 2021-03-12
Archive/job outputs

pagrubel:
This pull request places scripts in bee_workdir/<workflow id>/<task_name>


rstyd (approved):
Looks good to me! I ran it, and everything is where it should be.

PR #307, opened 2021-03-19
Generate metadata for task, and pass task object in submit and job qu…

pagrubel:
…eues.

Generated Task Metadata (dictionary) includes:
  - job_id
  - hostname
  - container runtime (if container is specified for task)
  - hash of container file  (if container is specified for task)

The queues (submit_que and job_queue) now have the task object so all task information is available.

### DIFF
--- a/-
+++ b/-
@@ -1,1 +1,0 @@
-fg-fey1.lanl.gov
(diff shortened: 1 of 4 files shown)


rstyd (approved):
Looks good to me!

PR #323, opened 2021-04-19
Multi TM Communication + Simple Worker code

jtronge:
This adds the ability for multiple Task Managers to communicate with the WFM and also adds a SimpleWorker class for systems without Slurm/LSF.


jtronge:
I've decided to close out this PR in favor of making some smaller PRs that won't conflict too much with existing code.

PR #332, opened 2021-05-20
Allow for no DockerRequirements

qwofford:
Currently build interface gives error if no DockerRequirements are given. Don't do that.


pagrubel (approved):
This looks fine.

Issue #329, opened 2021-05-17
build: Builder should allow flexible container archive directory


qwofford:
Right now, builder assumes `bee_workdir` is the base for container archive tarballs. Since these containers can be >20GB, we would like to be able to control where the container archive exists separately from the bee_workdir.

Suggested resolution is, on the first launch of `bee_interface.py`, check for config location of container_archive, if not exist, set up defuault and write to config. User can go back in and change later if desired.

Note to self, don't forget documentation!

pagrubel:
Need to make sure this is complete including documentation then close it. Adding to Public Release and assinging myself

pagrubel:
Closing this since the user can change the container archive location in bee.conf

PR #337, opened 2021-05-26
Make container archive configurable 

qwofford:
Checks the builder section of the config file for a `container_archive` key. If no such key exists, a default of `~/.beeflow/container_archive` is chosen and written to the config file with key `container_archive`.

Addresses #329


pagrubel (approved):
Looks good, pulled the container and tried again give the message that it is already there. Also the configuration was automatically fixed and the tarred image saved in the specified directory.

PR #343, opened 2021-06-15
Requirements Refactor

### LINKED ISSUE #341: Workflow Requirements Parsing and Improvement
The CWL parser needs to support parsing of `cwl_utils` requirement classes. Since each requirement class is represented by its own Python class, a parsing solution is needed to handle each requirement class specially. To simplify parsing, a suitable, minimal subset of requirement classes should be chosen to be supported.

`Hint` and `Requirement` data structured should also be refactored to hold a `namedtuple` of `(requirement class, key-value pairs)` to allow for cleaner database representation and easier implementation of requirement handling in the task manager and container runtime modules.

Boogie3D:
This PR makes the following changes:
* ~~Parser supports parsing of both hints and requirements~~
* `WorkflowInterface.create_requirement()` and `WorkflowInterface.create_hint()` now take two arguments
* `Requirement` and `Hint` classes now have two attributes: `class_` and `params`
  * `class_`: requirement class (`str`)
  * `params`: requirement parameters (`dict`)
* Neo4j now stores `Requirement` and `Hint` nodes with arbitrary properties instead of a single key-value pair
* Removed `BuildTask` from `wf_data.py`
* All tests in `test_wf_interface.py` are passing

Closes #341


pagrubel (commented):
Needed some fixes so our current example will work.

qwofford (approved):
Works for the test case I use for workflow validation. Thanks @Boogie3D !

qwofford (changes requested):
Let's revisit the Serious Python/PyPA and see if the CWL files in the parser2/ directory should be in src/beeflow/data.

pagrubel (approved):

PR #348, opened 2021-07-09
Workflow ID Patch

Boogie3D:
Simple patch fixing failure to propagate workflow ID to tasks
- `wf_interface.initialize_workflow()` now stores workflow ID in private attribute
- `test_wf_interface` now tests for propagation of workflow ID to tasks


qwofford (approved):
Works for me

pagrubel (approved):

PR #350, opened 2021-07-21
Simple worker

jtronge:
These changes add the SimpleWorker class that should allow for running BEE workflows without a batch scheduler. This is useful for testing and debugging purposes, and is also what I currently use for running workflows in the cloud. I imagine this could also be useful for running a workflow locally before running it on an HPC system.


jtronge:
@pagrubel You're right. I removed it.

pagrubel (changes requested):

pagrubel (approved):

PR #352, opened 2021-07-26
Generic CWL Parser

Boogie3D:
Adds generic CWL parser which currently supports the following CWL features:
- `Workflow` and `CommandLineTool` (parsing of individual `CommandLineTool` files without a corresponding `Workflow` file is currently not allowed)
  - Typed workflow inputs and outputs
  - Step `run` field to specify corresponding `CommandLineTool` file or inline `CommandLineTool` definition
- All primitive types including `File` and `Directory`
  - Does not support complex types such as `record` or `array`
- JSON and YAML job parameter files
- Workflow hints/requirements
- Step hints/requirements
- List of step outputs
- Step input type, optional marker `?`
- Step default value
- Step input binding
  - Prefix
  - Position
  - ValueFrom (only supports string concatentation, parameter substitution of form `$(inputs.foo)`, and `self.path` special variable for absolute path of input file/directory
- Step output binding
  - `glob` field
  - Supports parameter substitution of form `$(inputs.foo)` in `glob` field

Modifies following data structures to `wf_data.py`:
- `OutputParameter`
- `StepInput`
- `StepOutput`
- `Workflow`
  - `inputs` and `outputs` stored in `list`s
  - `__eq__()` now sorts `inputs` and `outputs` by `id` before comparison
- `Task`
  - Added `stdout` field to specify `stdout` capture
  - Removed `subworkflow` field since not used
  - Updated `__repr__()`
  - `construct_command()` now implemented as `command` property
    - Constructs and returns `Task` command from `inputs` as a list
    - Raises `ValueError` if any inputs not satisfied
  - `inputs` and `outputs` stored in `list`s
  - `__eq__()` now sorts `inputs` and `outputs` by `id` before comparison

Modifies `wf_interface.py`:
- `__init__()` now establishes database connection (no longer `initialize_workflow()`)
- `execute_task()` now populates initial task inputs from workflow input values
- `finalize_task()` now copies task output values to corresponding dependent task inputs or workflow outputs
- `Added `workflow_id` property
  - Grabs `workflow_id` from graph database if not stored in protected member `_workflow_id` (i.e. new WFI instance after workflow initialization)
- Added `get_workflow_outputs()` to retrieve workflow outputs
- Added `get_task_input()` to retrieve a task input
- Added `set_task_input()` to set a task input
- Added `get_task_output()` to set a task output
- Added `evaluate_expression()` to evaluate a task input ValueFrom expression or output glob expression
- Removed `get_workflow_id()` as now redundant
- Removed `get_subworkflow()`
- Removed `existing` parameter from `initialize_workflow()`
- `initialize_workflow()` raises `RuntimeError` if attempting to re-initialize existing workflow

Modifies `gdb_interface.py`, `gdb_driver.py`, `neo4j_driver.py`:
- Removed anything related to `subworkflow`
- `Task` reconstruction updated for new data structures

Modifies `neo4j_cypher.py`:
- Removed any functions related to `subworkflow`
- `foo_id` properties renamed to simply `id`
  - e.g., `task_id -> id`, `input_id -> id`, etc.
- Dependency resolution query updated for new input/output nodes
- Unique constraint is now on task ID instead of name

Modifies `test_wf_interface.py`:
- All tests passing for latest branch
- Added missing docstrings
- Removed `test_get_subworkflow()`


pagrubel:
@rstyd The client needs to use the new parser for "authentic" cwl files.
@Boogie3D can you point us to the cwl file we should use as a test for the clamr example?

pagrubel:
sorry not the client, but the wf_manager needs to use cwl files to submit a workflow

qwofford:
How do I test this?

Boogie3D:
> How do I test this?

The [README](https://github.com/lanl/BEE_Private/blob/develop/src/beeflow/common/parser/README.md) provides instructions on how to use the parser. There aren't any unit tests for it right now. Should I make some using the CLAMR example?

qwofford:
I think unit tests are a good idea but not the priority for me at the moment. 

I'm wrapping up the build interface integration so it's most useful to test with a full workflow run on Fog or Darwin using the client. I'm more interested in seeing the parser functioning for a specific CWL file from start to finish, rather than seeing the parser work for a variety of cases. While parsing a variety of workflows is certainly important for the project more generally, demonstrating the build interface integration only requires an "easy" CWL case, with minor modifications to the DockerRequirement section of tasks.

Toward that end, I think we can just wait until Rusty pushes his changes to this branch on Monday. After this is merged into develop, mine should fall into place in short order.

mcpherson:
Going to move old parsers (`parse_clamr.py` and `parse_cwl.py`) and tool (`dump_cwl.py`) to `deprecated/` so it's clear that there is only one parser. We can delete `deprecated` when the PR is merged.

pagrubel:
Why would we need to even keep them before merging?

mcpherson:
Didn't know if @rstyd or @qwofford were doing A/B or anything similar. I can delete 'em now if nobody is using them (in the context of this branch).

PR #356, opened 2021-08-05
Add workflow profiling code

jtronge:
This is another smaller PR which adds a workflow profiling script. I'm going to mark this as WIP for right now, until I can add this code to the workflow manager and add some configuration options to the `bee.conf`. Basically this is useful for running experiments with BEE workflows and then capturing task runtimes, output results from the scheduler, etc. I've used this for experiments in a couple of our paper submissions and I could see this being useful for others who want to measure workflow performance. Once this PR goes through I plan to open another with the cloud launching code.

### DIFF
--- a/src/beeflow/common/wf_profiler.py
+++ b/src/beeflow/common/wf_profiler.py
@@ -1,0 +1,45 @@
+"""Workflow profiling code."""
+import json
+import time
+import os
+
+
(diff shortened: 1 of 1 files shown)


jtronge:
I'm going to close this PR without merging for right now. I plan on including this in with the MPI code that I'm working on.

Issue #361, opened 2021-08-23
Task output and scripts in incorrect directory for LSF


pagrubel:
LSF worker does not place the task files in the correct directory so they aren't being archived either. I have a fix and will submit a PR

pagrubel:
Fixed in PR #362

PR #362, opened 2021-08-24
Fix path for scripts and outputs for task/job.

pagrubel:
Simple fix for LSF so that the step job scripts and outputs will be in the proper path and archived.


pagrubel:
Fixes issue #361

PR #370, opened 2021-08-26
make logic right for preferring hints over requirements.

### LINKED ISSUE #365: build: Validate requirements and hints are handlled appropriately.
Fix if they aren't. Looks like scenario with both hints and requirements might prefer hints over requirements, which is wrong.

https://github.com/lanl/BEE_Private/blob/761dbc6aad9fd1d96e3ecb84e3fe1f636bb8aa76/src/beeflow/common/build/container_drivers.py#L183-L187

This will need to be corrected throughout the file. Potential code movement opportunity.

qwofford:
Address issue #365

### DIFF
--- a/src/beeflow/common/build/container_drivers.py
+++ b/src/beeflow/common/build/container_drivers.py
@@ -181,7 +181,7 @@ class CharliecloudBuildDriver(ContainerBuildDriver):
             req_addr = None
 
         # Prefer requirements over hints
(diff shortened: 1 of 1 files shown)


qwofford:
Tested with `src/beeflow/data/cwl/bee_workflows/clamr-ffmpeg/fog/no_yaml/cf2.cwl`

PR #387, opened 2021-11-02
Set Input Type After Evaluating Expression

Boogie3D:
Patch adds code to change input type in the graph database:
- Added `set_task_input_type()` method to `GraphDatabaseInterface`, `GraphDatabaseDriver`, and `Neo4jDriver`
  - Called by `WorkflowInterface.evaluate_expression()` before setting input value when called with `output=False`
- Added `set_task_input_type()` function to `neo4j_cypher.py`


pagrubel (approved):

rstyd (approved):
Looks good to me

PR #388, opened 2021-11-02
Revert "Generic CWL Parser"

pagrubel:
Reverts lanl/BEE_Private#352


pagrubel:
Decided to fix rather than revert.

PR #390, opened 2021-11-05
Clean up bee_flow cwl files

pagrubel:


pagrubel:
This PR is ready to be merged. I've checked all the clamr workflows on Summit,  Darwin and Fog.

Although I think we can delete src/beeflow/data/parser_example/ but would like a nod from @Boogie3D

pagrubel:
I removed parser_example after confirming with Steven. If you think the orgainization of the cwl files should be something other than what I have done, make a suggestion.

Boogie3D (approved):

PR #406, opened 2021-12-02
Make ch-builder2tar use ch-image by default

jtronge:
`ch-builder2tar` will use Docker to export containers if it's installed on a given system. We have to explicitly add `-b ch-image` to get it to export the container from Charliecloud. I noticed this when using a workflow with `dockerPull` on my machine.

### DIFF
--- a/src/beeflow/common/build/container_drivers.py
+++ b/src/beeflow/common/build/container_drivers.py
@@ -220,7 +220,7 @@ class CharliecloudBuildDriver(ContainerBuildDriver):
                 pass
 
         # Out of excuses. Pull the image.
-        cmd = (f'ch-image pull {addr} && ch-builder2tar {ch_build_addr} {self.container_archive}'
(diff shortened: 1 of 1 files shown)


pagrubel:
Do we want to hard code this option? I thought ch-image was default.

jtronge:
I think it's okay to hard code this, since it's in the Charliecloud driver and we want to use ch-image here.

Maybe @rstyd would know, is this how `ch-builder2tar` is supposed to work? Maybe my system is configured incorrectly.

pagrubel:
I believe there is a way to set your environment to use ch-image, but I'm okay with this as long as there is no way someone would want to use docker or some other builder.

jtronge:
I'm closing this PR based on the meeting on 12/6/21. According to Rusty `ch-builder2tar` will be deprecated soon and using `CH_BUILDER` fixes this particular issue.

Issue #405, opened 2021-12-01
BeeConfig Instances


jtronge:
When loading up components of BEE, multiple instances of the BeeConfig class are getting created. Because of this, configuration within a single component is not the same.

As an example, in the Task Manager, one BeeConfig instance gets created in that particular script, then when the container runtime code is imported, another instance gets created in that code, then more instances are created in other submodules (in different parts of the `common/` package). If somebody needs to use a different bee.conf file for some reason, I've noticed that some parts of the code load the correct config file, while other parts load the old configuration file and then some parts don't work as they should.

I'm thinking that we may want to turn the BeeConfig class into some sort of "singleton" object or implement configuration loading at a module level, so that configuration is consistent. So for example, we'd want to initialize the config in only one place (the `wf_manager.py` script, the `task_manager.py` script, etc.) and then other modules would only need to import the BeeConfig module to get the correct config values.

pagrubel:
I'm still a little confused. When you say "we'd want to initialize the config in only one place (the wf_manager.py script, the task_manager.py script, etc.) " Do you mean in only one place per each module (task manager, wf manager etc.)?

jtronge:
I think we'd want to do these once per process (so once per task manager, wf manager, scheduler, etc) rather than per module. So we'd only have this code chunk once in the main script file, rather than in each module file: 

```
if len(sys.argv) >= 2:
    bc = BeeConfig(userconfig=sys.argv[1])
else:
    bc = BeeConfig()
```
The big issue right now is that when we load the task manager or the wf manager, these objects are getting created in different modules throughout the `common/` directory, and the same chunk of code above is copied to many of those modules. Sometimes however they aren't loading the right configuration file. So if I change some config options under the `[charliecloud]` configuration, one module might see that change, but another won't if they don't load the config in exactly the same way.

To avoid this, I'm thinking we should only run that code once, in each main script that gets run.
This shouldn't change the way that BEE is configured (so if we want the a different configuration per wf manager or task manager, etc. that would still work).

pagrubel:
@jtronge you mentioned in the meeting that this causes bugs. Should we add the bug label?

pagrubel:
@jtronge Is work on this issue completed?

jtronge:
Yes this can be closed.

Issue #403, opened 2021-11-23
Replace job templates with configurations in bee.conf


pagrubel:
I think we should get rid of the job template and make a way to specify setups  in bee.conf for the particular worker. That will work for a cluster, but ultimately it should be per job so we will have to decide to support some through BEE unique cwl attributes.

Examples of how we have been using the job templates were to specify an account for the runs (required on Summit and other systems), specify a partition or other constraint for the system, even to test on a certain set of nodes etc.

pagrubel:
First step: Use bee.conf for specifying cluster wide congiguration for #SBATCH or $BSUB directives
then add option for steps in CWL.

pagrubel:
We'd like to see this done before the release to prevent users having to change their workflow.

pagrubel:
In PR #412 We require jinja job templates which give us flexibility for each step. However, it may still be nice to have some easier method for the users. Thinking on this.

pagrubel:
We are currently using jinja which allows us to do things per task such as MPI and we can add others

pagrubel:
We are thinking or reworking the use of the jinja file so keeping this issue open.

pagrubel:
I'm closing this issue. I think that we will need issues for incorporating CWL extensions for items that we presently use jinja files for.

PR #412, opened 2021-12-21
Initial MPI Support

jtronge:
This is an initial PR for adding MPI support with BEE. I've made a number of other changes as well and some of the code is not implemented in all of the workers, so I'm marking this initially as a WIP.

The main addition this adds is support for a new `beeflow:MPIRequirement` hint/requirement. Let me know if you think I should call it something different. To do this I've modified the Slurm worker code and some of the container runtime code to use `Command` objects when building the submission script.

You can take a look at the workflow in `src/beeflow/data/cwl/bee_workflows/comd-mpi` for an example of some parts of this requirement.


src/beeflow/common/config_driver.py
```diff
@@ -5,7 +5,19 @@
 import platform
 
 
-class BeeConfig:
+def BeeConfig(*pargs, **kwargs):
```
jtronge: I think this should fix [Issue #405](https://github.com/lanl/BEE_Private/issues/405)

src/beeflow/data/cwl/bee_workflows/nwchem-mpi/README.md
```diff
@@ -0,0 +1,44 @@
+## NWChem MPI Workflow
```
jtronge, line 1: I added instructions for running this `nwchem-mpi` workflow in this README here.

pagrubel:
This should satisfy #426 .

pagrubel:
@jtronge So I believe these are the flags that will work for LSF (Summit):

[#jsrun](https://mattermost.energy.gov/beecharliecloud/messages/@jtronge#) -r 1 -c 1 
jsrun --rs_per_host 1 mkdir -p /var/tmp/<username>/beeflow_deployed_containers
jsrun --rs_per_host 1 ch-tar2dir <tarball>  /var/tmp/<username>/beeflow_deployed_containers

and for running there are many options so that should be flexible
I have a co worker that uses -n 256
but I'd say for the example that you had for nwchem it would be the following:

jsrun --rs_per_host 1 --tasks_per_rs 2 ch-run ...
jsrun -r 1 -c 1

jtronge:
I changed the CoMD workflow (in `src/beeflow/data/cwl/bee_workflows/comd-mpi/`) to be a multi-file workflow, so it should work with the parser. This example runs with 8 nodes, 1 task per node. The container file can be built with `ch-image build --force -f Dockerfile.comd-x86_64 -t comd .` from the `src/beeflow/data/dockerfiles` directory.

jtronge:
I'm thinking that to ensure that we can run jobs with many different possible options (MPI version, etc.) it might be better to put the script generation code into some sort of template. Putting all this logic in the worker code itself might not be the best option. I think I might try to make use of the existing template file code within the workers.

pagrubel:
I was going to eliminate the job template in favor of system settings for the #SBATCH or equivalent LSF directives. See issue #403

jtronge:
I wonder if we would be able to discuss this more at the meeting tomorrow? My original plan was to use a separate yaml file, similar to what CWL uses for their extension, but with more options. I'm realizing though, the problem with this set up is that it leads to more complex code in the workers (we have to add checks for each MPI requirement in each worker).
There are also other system-specific options that might need to be added on certain systems (for example the `--mpi=pmix_v3` that we needed for Darwin), and these would have to be added somehow to the worker code or to the command-line code.
This could also cause problems for when or if BEE eventually supports multiple task managers that each might work with a different resource manager.

So, instead, I'm wondering if it would be possible to generate a submission script using a system-specific template module that takes the commands generated by the container runtime, as well as specific hints/requirements as input. I believe this would allow for system-specific options to be taken into account, while still ensuring that workflows are reasonably portable between systems.

jtronge:
Here's an example of what I mean by a system-specific template module. This is using Jinja, which is what I also used in the cloud launching code, and is already a BEE dependency.

```
#!/bin/bash
#SBATCH --job-name={{ task_name }}-{{ task_id }}
#SBATCH --output={{ workflow_path }}/{{ task_name }}-{{ task_id }}.out
#SBATCH --error={{ workflow_path }}/{{ task_name }}-{{ task_id }}.err
{% if 'nodes' in task_info %}
#SBATCH -N {{ task_info['nodes'] }}
{% endif %}

{% for command in commands %}
{% if command.block is not none %}
{{ command.block }}
{% else %}
srun {{ command.argv|join(' ') }}
{% endif %}
{% endfor %}
```

src/beeflow/common/build_interfaces.py
```diff
@@ -17,59 +17,67 @@
...
 
-bee_workdir = bc.userconfig.get('DEFAULT', 'bee_workdir')
-handler = bee_logging.save_log(bee_workdir=bee_workdir, log=log, logfile='build_interface.log')
-
-try:
-    task = arg2task(my_args)
+def build_main(bc, task):
+    """Main build code."""
     # The build driver treats Hint and Requirement objects as Dicts.
+    task = task.copy()
```
pagrubel: @jtronge can you explain why you are making a copy here? Would it be more understandable if the copy had a new name such as local_task

pagrubel:
@jtronge This PR looks great and our basic example works fine. 
I was able to run mpi jobs on two different platforms using copyContainer, dockerFile with containerName and dockerPull. I'll probably add some dockerfiles that I need to build for darwin which has slurm with no pmix support. 

I especially like how you have moved some of the individual worker code into worker.py by using the jinja templates.   I still have some concern for users on managing jinja files especially in the future as we add more BEE CWL extensions.
Rusty and I went through the code together and have some questions for you, comments to follow.

src/beeflow/common/crt_drivers.py
```diff
@@ -241,21 +226,20 @@ class SingularityDriver(ContainerRuntimeDriver):
...
-                        ' ', command,
-                        ])
-                    docker = True
-            if not docker:
-                text = command
-        return text
+            hints = dict(task.hints)
+            # I'm not sure if this should check for copyContainer here or not
+            try:
+                img = hints['DockerRequirement']['dockerImageId']
```
pagrubel:
```suggestion
                img = hints['DockerRequirement']['copyContainer']
```

src/beeflow/common/crt_drivers.py
```diff
@@ -241,21 +226,20 @@ class SingularityDriver(ContainerRuntimeDriver):
...
-                    text = ''.join([
-                        'singularity exec ', hint.params["dockerImageId"],
-                        ' ', command,
-                        ])
-                    docker = True
-            if not docker:
-                text = command
-        return text
+            hints = dict(task.hints)
+            # I'm not sure if this should check for copyContainer here or not
```
pagrubel:
```suggestion
```

pagrubel (changes requested):

src/beeflow/common/build_interfaces.py
jtronge: I made a copy here because the build driver is modifying the task object itself. Otherwise it affects other code in the task manager. I updated the name to be local_task

src/beeflow/common/crt_drivers.py
jtronge: I updated this. I just wasn't sure originally whether I should do that because I don't use the singularity driver

jtronge:
Thanks Pat. I think I made the changes. Let me know if you notice anything else.

pagrubel:
I added a working singularity example, although it is minimal.
I also added dockerfiles and instructions to build a container for the comd example with pmix support in the container. If slurm is not installed with pmix support the support needs to be in the container.

src/beeflow/common/build_interfaces.py
```diff
@@ -17,59 +17,67 @@
...
-log.info('CharliecloudBuildDriver initialized')
-
-# Deque the next build instruction until empty or reach terminal case
-# ...catch the case where no build instructions are required.
-try:
-    build_op, op_name, op_priority, op_terminal = builder.exec_list.pop(0)
-except IndexError:
-    log.info('No build instructions provided. Assuming this is ok...')
-    build_op, op_name, op_priority, op_terminal = None, None, None, True
+    local_task = task.copy()
```
rstyd, line 24: I was wondering what exactly does this copy do? Is it just to prevent modifying the original task when we turn the hints and requirements into dictionaries?

src/beeflow/wf_manager.py
```diff
@@ -630,6 +662,16 @@ def put(self):
         return resp
 
 
+
+@flask_app.route('/bee_wfm/v1/status/<string:wf_id>')
```
rstyd, line 666: What exactly is this route for? Looking at it I'm thinking it's part of the profiling, but I wanted to be sure.

jtronge: I originally added this for the beeswarm client. Basically I used it to check for the status of a submitted workflow. I think there's another endpoint for this, `JobActions.get()` but it wasn't working for me. I can probably remove this for right now.

src/beeflow/common/build_interfaces.py
jtronge: Yeah, so to kind of give some background on this, when testing I was having a lot of trouble debugging the build driver. It seemed to fail arbitrarily and I realized that it was getting called through a `subprocess.run` call. A couple times I wasn't able to figure out what went wrong, because Charliecloud would fail or something else would happen. I think the log output and specific failures were just getting lost somehow.

I thought one way to possibly fix this issue was to invoke the build driver directly in Python, instead of using a `subprocess.run` call. That way the log would be inherited from the task manager, and we might be able to better note when errors occur. This change is in `src/beeflow/task_manager.py:160`. Basically I do a `from beeflow.common.build_interfaces import build_main` and then call build_main directly. I though this made sense since, at least for right now, the build interface seems to be a sort of sub-step or sub-component of the task manager.

Anyway, once I did that, I was able to get more output back from the build driver, but then I realized that it was modifying tasks in place. This is why I added the `task.copy()` function there, because otherwise it will mess with the other code in the task manager. There's definitely a better way to do this. I think we'll probably need to make some changes here in the future, but I'm just not overly familiar all of the code in the build driver yet.

Sorry for so much text, just thought my changes to the build interface needed a little more context.

src/beeflow/wf_manager.py
rstyd: Ah, okay yeah I've recently made some changes to that endpoint which will be brought in with the new CLI client PR.

src/beeflow/common/build_interfaces.py
rstyd: All the background info really helps so no need to apologize! That makes sense to me. I think the change of not using subprocess like that is good since it was a bit awkward. Might need a bit more refactoring around it in the future, but it's a solid change.

rstyd:
I approve!

pagrubel (approved):
Thanks for answering our questions Jake.

PR #413, opened 2022-01-05
Fix poetry

pagrubel:


pagrubel:
@Boogie3D Rusty(@rstyd) fixed the poetry problem by upgrading cwl_utils to .11 We discussed upgrading to the latest which required a few more changes. This PR makes those changes. I was able to run the standard clamr workflow and the exercise in the README with these changes, but was wondering if 
parse_clamr.py is still needed. Also, you may want to verify any tests that you use.

Boogie3D (approved):

PR #438, opened 2022-03-21
Short Workflow ID Support

### LINKED ISSUE #422: Change Workflow ID
From Public Release discussion (Rusty):
We had discussed changing the workflow IDs to make them more readable. Currently we’re using UUIDs.
I did a little bit of googling and found this page with a nice discussion about IDs: https://gnugat.github.io/2018/06/15/short-identifier.html.
I like their idea of using the first few characters of a SHA256 hash the same way that git does.

From previous discussion(Tim):
I think the WorkflowID (WfID?) should be something meaningful and easy for a human to interpret. Alternatively, if the WfID is something ugly like a git commit hash, BEE should implement a way of mapping that to a human-friendly representation. Some random ideas for WfID:

name-date where name is supplied by the user via the client and date is provided by BEE in the format YYYYMMDDHHMM. An example would be VASP-SnO-202005061201.
UUID where UUID is created by BEE using python's uuid.uuid4() function. BEE would allow a user to refer to this workflow in the same manner as git does:

Boogie3D:
Workflow IDs are generated using UUID, which in their hex representation are 32 characters long, making them inconvenient and unwieldy for users. This PR adds support for shortened workflow IDs and resolves #422.

Changes:
* Workflow ID generation refactored to `Workflow.generate_workflow_id()` method
* Workflow name removed from ID, leaving only UUID
* Client supports user-side shortened workflow IDs
  * Full-length IDs still stored in database
  * Shortened 6-character IDs presented to user by default
  * Checks for short ID collisions when workflows are added/re-executed, increments short ID length if any detected
  * All client functions that accept workflow IDs as user input can now take partial workflow IDs, matching them against full IDs from job list
    * `start_workflow()`
    * `query_workflow()`
    * `cancel_workflow()`
    * `pause_workflow()`
    * `resume_workflow()`
    * `copy_workflow()`
  * Client `list_workflows()` function now displays shortened workflow IDs

### DIFF
--- a/src/beeflow/client/client.py
+++ b/src/beeflow/client/client.py
@@ -56,6 +56,7 @@ def safe_input(type):
             # Cast it to the specified type
             answer = type(answer)
             break
+            # Put logs in main bee_workdir log directory /logs
         except ValueError:
             print(f"Error {answer} is not a valid option")
     return answer
@@ -63,6 +64,7 @@ def safe_input(type):
 # Check short workflow IDs for colliions, increase short ID
 # length if any detected
 def check_short_id_collision():
+    global short_id_len
     resp = requests.get(_url())
     if resp.status_code != requests.codes.okay:
         print(f"Returned {resp.status_code}")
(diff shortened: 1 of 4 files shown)


pagrubel:
Tested this on fog and it worked fine. I even changed tried having multiple workflows starting with the same first digits (by manually changing directory name). All it gives is an ambiguous warning which is fine for now. I doubt we will come across this, but we could at least list the choices at some point. I'll try this out on summit too.

pagrubel (approved):

rstyd (approved):
LGTM!

Issue #415, opened 2022-01-24
Update for Charliecloud v 0.26/0.27


pagrubel:
ch-build is deprecated in v 0.26 of charliecloud and will be gone in 0.27:
We need to use the newer commands and only support Charliecloud 0.26 +
Here are the release notes for context:

Subject: Charliecloud 0.26 released
 
Dear Charliecloud community:

We are pleased to announce the release of Charliecloud version 0.26.

Highlights:

  * “ch-run” can now mount SquashFS images internally, eliminating the separate mount and unmount using “ch-mount” and “ch-unmount” (#1081). These two scripts have been removed, but separate mount/unmount is still available by invoking “squashfuse” or “squashfuse_ll” and “fusermount -u” directly. Note this feature requires linking with a bleeding-edge version of SquashFUSE.

  * New script “ch-convert” that can convert between any supported image format (#698). This directly replaces “ch-build2dir”, “ch-builder2squash”, “ch-builder2tar”, “ch-dir2squash”, and “ch-tar2dir”, which provided a subset of its functionality. “ch-pull2-dir” and “ch-pull2tar” are also now deprecated, replaced by invoking the image builder directly (e.g., “ch-image” or Docker) followed by “ch-convert”. These seven deprecated scripts are scheduled for removal in 0.27.

  * “ch-build” is now deprecated and scheduled for removal in 0.27 (#1225). The replacement is to invoke the image builder directly; we have added a FAQ to help with Docker.

  * The “ch-image” default storage directory path has been changed from /var/tmp/$USER/ch-image to /var/tmp/$USER.ch/ch-image (#1160). In most cases, an existing storage directory at the default path will be moved automatically.

pagrubel:
Closing this since #444 fixed it Note: Charlicloud v 0.27 or later is now a requirement

PR #444, opened 2022-04-21
Update ch-tar2dir usage with ch-convert

rstyd:
Addresses #415


src/beeflow/common/build/container_drivers.py
```diff
@@ -351,11 +352,8 @@ def dockerImport(self, param_import=None):
             import_input_path = task_import
 
         # Pull the image.
-        dir_name,ext = os.path.splitext(os.path.basename(import_input_path))
-        while ext:
-             dir_name,ext = os.path.splitext(dir_name)
-
-        cmd = (f'ch-convert {import_input_path} {self.deployed_image_root}/{dir_name}')
+        file_name = crt_driver.get_ccname(import_input_path)
```
pagrubel, line 355: @rstyd I used this method in the Charliecloud driver for getting the basic image name from the path. Should we put it in a more general place?

pagrubel:
@rstyd I think I fixed the rest of things that make us compatible with Charliecoud 0.27 but now you should do a review

rstyd:
Looks good to me

pagrubel (approved):

Issue #440, opened 2022-03-24
Configuration defaults bee.conf


pagrubel:
We need to insure that bee.conf with all defaults needed is created in one location. 
One location in the code that is.

pagrubel:
Closing this it was solved with #477

PR #447, opened 2022-05-05
Config Improvements

jtronge:
This is an initial PR that should update BEE to fix issues #440 and #405.


jtronge:
I think the validation code in the `config_driver.py` should be complete. You should be able to run `beeflow-cfg --validate [fname]` and it will validate your configuration file, telling you if something is missing or not working.

The updates I've made to the BeeConfig object remove the userconfig/sysconfig property. Rather than directly accessing the userconfig/sysconfig property, I've added a single `get(section, name)` method to be used for accessing config variables. My idea here is that this will allow for better validation of variables in the config, and will also allow for userconfig/sysconfig to be implemented in the config driver later on, without having to change code all throughout beeflow. The next step is to change all calls that access the old userconfig property, to use the new get method. I started with the task manager script, but I haven't made any changes elsewhere yet.

I've also started some code that should eventually help users generate the `bee.conf`, but this is not quite functional yet. You can try it out by running `beeflow-cfg --new [fname]`, but again it will likely not run correctly. This uses the same code that I've used for validation.

jtronge:
I think most of the code should be working now with the new BeeConfig interface. I was able to run the clamr example without running into any errors.

The config generation code needs a little bit more testing and I need to cleanup some left-over debug code/comments, but I think this PR will soon be ready for review.

src/beeflow/common/config_driver.py
```diff
@@ -193,3 +138,321 @@ def resolve_path(relative_path):
...
+    if template not in template_files:
+        raise NotImplementedError(f'generation of template file "{template}" is not implemented')
+    fname = template_files[template]
+    # I don't like how this is done, but there seems to be some difficulties
+    # with using the pkgutil.get_data() method and the importlib.resources
+    # module.
+    common = os.path.dirname(__file__)
+    beeflow = os.path.dirname(common)
+    file_path = join_path(beeflow, 'data', 'job_templates', fname)
+    shutil.copy(file_path, path)
```
jtronge, line 249: Here I'm trying to copy an example Jinja template over to a location given by the user. I'm not sure if this is the best way to do this, or if this will work once beeflow is distributed with a package.

jtronge:
I've added a number of test cases for my config validation code, which you should be able to run with `pytest src/beeflow/tests/test_config_validator.py`. Also, it would be good run `beeflow-cfg --validate [bee.conf path]` to ensure that existing configs are considered valid (if they aren't it will list the invalid options).

Also, I added another option to the configuration script, called with `beeflow-cfg --info`, which will dump a list of all possible config options and a short description of each for the user.

src/beeflow/common/config_driver.py
```diff
@@ -193,3 +138,322 @@ def resolve_path(relative_path):
...
+# case it's being used to store a special 'init' function that can be used to
+# initialize the parameter when a user is generating a new configuration.
+VALIDATOR.option('task_manager', 'job_template', required=True,
+                 info='job template to use for generating submission scripts',
+                 validator=validate_file,
+                 attrs={'init': job_template_init})
+# Charliecloud (depends on task_manager::container_runtime == Charliecloud)
+HOME_DIR = os.path.expanduser('~/')
+VALIDATOR.section('charliecloud', info='Charliecloud configuration section.',
+                  depends_on=('task_manager', 'container_runtime', 'Charliecloud'))
```
jtronge, line 326: I wanted to note that I've made the validation so that some sections explicitly depend on certain values. For instance, the `[charliecloud]` section depends on `task_manager::container_runtime` being set to `Charliecloud`. This ensures that if a user includes the charliecloud section, but doesn't set the container_runtime right, then an error will be thrown. This works only at the section level right now, not with individual options.

pagrubel:
I'm trying to run this as a brand new user. Here are some ideas:
All logs should automatically be in the bee_workdir/logs. I'm not sure you would want logs all over the place. We should archive the logs, so that there is a record for provenance and debugging.

So don't choose anything for logs but you can have a choice for bee_workdir
What else needs to go in bee_workdir? We all need to think about this.

The container_archive directory can be placed in a different path since often the home directory has tight limits and containers can become very large. We should probably have some name to the directory that is common though.Like bee_container_archive. The user could give a path and we have that as a subdirectory.


To be consistent beeflow-cfg should probably be bee_config and be more like bee_client. (Although we could defer this if it would take too long)

Just typing bee_client gives help
bee_config list could list all the current options

Do you want to write a default job template to "... should probably only be asked if one doesn't exist or a warning should happen before overwriting it.

Ok this seems petty but I really did not know that NEW or VALIDATE meant a file path. It may be a good thing that my brain is foggy, but I think we need to be specific.

something like this:
--validate CONFIG_FILEPATH, -v CONFIG_FILEPATH 
                            validate configuration file
--new CONFIG_FILEPATH, -n CONFIG_FILEPATH
                       create a new bee configuration file

pagrubel:
It would be nice to have a list (or name it whatever) to print out the current configuration

bee_cfg list
It could include the file path, incase someone wants to edit it

pagrubel:
validate could validate ~/.config/beeflow/bee.conf if no filepath was given making FILE optional

pagrubel:
job_template: ~/.config/submit.jinja
Do you want to write a default job template to "/users/pagrubel/.config/submit.jinja"? [y/n] y
If I pick this option it will overwrite the file if it exist, you may want to ask if they want to overwrite

pagrubel:
Many things that were default are not the same anymore:
slurm socket
container_archive
use_archive

and the defaults that are used are not put in the config file now, so it isn't easy to figure out what is going on. I have no idea what ports I'm using and maybe a user doesn't need to know but it would be nice to be able to look at what they are using.

bee.conf goes in the archive for provenance and we need to populate it with all the defaults

pagrubel:
cls.CONFIG = VALIDATOR.validate(config)
  File "/yellow/users/pagrubel/BEE/BEE_Private/src/beeflow/common/config_validator.py", line 45, in validate
    raise ValueError(
ValueError: DEFAULT::bee_workdir: path "/users/pagrubel/.beeflow" does not exist

So create it

pagrubel:
present default for container_archive is $HOME now should be bee_workdir/container_archive
slurm socket is /tmp/slurm.sock so will we be stepping on each other
previously it was 
```
[slurmrestd]
slurm_socket = /tmp/slurm_pagrubel_127.sock
```

Boogie3D (approved):

pagrubel (approved):
Looks good, also nice path to get a new user up and running with a default bee.conf and jinja file. There is an issue running on summit but we can deal with that separately.

Issue #446, opened 2022-05-04
Builder does not process image files with extension tgz


pagrubel:
Although the dockerRequirement in the cwl was
copyContainer: "/usr/projects/beedev/clamr/clamr-toss.tgz"

Apparently it was copied to the container_archive directory as clamr-toss.tar.gz

INFO: Expecting container at /yellow/users/pagrubel/.beeflow/container_archive/clamr-toss.tar.gz. Ready to deploy and run.

pagrubel:
Charliecloud ch-convert:  Output tarballs are always gzipped and must end in .tar.gz; input tarballs can be any compression acceptable to tar(1).
FYI

pagrubel:
PR #448 solved this issue.

PR #448, opened 2022-05-05
Issue446/allow tgz

pagrubel:
This PR fixes the problem with using copyContainer for an image tar with extension .tgz

After preprocessing DockerRequirements, the builder places an image in the container_archive and that is always a tar.gz file. So the run scripts will use that copy.


pagrubel:
This PR fixes #446

pagrubel:
This PR fixes issue #446

pagrubel:
@jtronge Problems you found should be fixed now. Comment here if you see any others.

jtronge:
I tried this with two comd workflows, one using `copyContainer` and the other using `dockerPull`. Both worked fine.

jtronge (approved):

PR #457, opened 2022-05-18
Chicoma example files

pagrubel:
This PR adds cwl example and dockerfile for the clamr example that works on the standard partition of Chicoma.


rstyd:
LGTM

Issue #460, opened 2022-05-25
Jinja Submit Template Lmod 


rstyd:
The jinja submit template is currently not working correctly with lmod. It uses `srun` before each command. 
This is an issue since one of the commands in the pre-command  in the container runtime driver can be `module load` depending on the environment in which the workflow is running. Using `srun` on `module load` doesn't work so we'll need to either nor srun any command which uses module load or add an additional module section to the container runtime driver which the jinja template which the jinja can add to the script without srun.

pagrubel:
@rstyd Can we close this issue now?

pagrubel:
Closing this as is fixed in #471

PR #461, opened 2022-05-25
Add hotfix for jinja submit on lmod systems

rstyd:
Addresses #460 with a temporary solution.


rstyd:
Going to modify this to put the module load step into a different command list then change the jinja template so that it won't srun those commands.

pagrubel:
Did #471 fix this can we close this

pagrubel:
Issue #471 was fixed in another PR

PR #462, opened 2022-05-25
Wfi/restart task

Boogie3D:
Make necessary changes to implement automatic checkpoint restart feature:
- Add `restart_task()` method to WorkflowInterface, GraphDatabaseInterface, GraphDatabaseDriver, Neo4jDriver
  - Can only be invoked on `Task` object with a `Hint` of class `beeflow:CheckpointRequirement`
     - If `hint.params["num_tries"]` = 0, returns `None`
     - Creates proprietary field in `hint.params` called `bee_checkpoint_file__`, which stores the checkpoint filename
  - Creates copy of failed `Task` node with new Task ID, state set to `RESTARTED`
  - Copies failed `Task` metadata to new, restarted `Task`
  - Appends a number in parentheses to the new `Task` name, which increments on each restart
- Modify `add_dependencies()` Cypher function to take `restarted_task` boolean parameter
  - Deletes `DEPENDS_ON` relationships pointing to failed `Task` node
  - Add `RESTARTED_FROM` relationship pointing from restarted `Task` node to failed `Task` node
  - Create `DEPENDS_ON` relationships from dependent `Task` nodes to restarted `Task` node
- Change `get_task_metadata()` method to not take `keys` as parameter, just return entire dictionary (minus `state`)
- Added sanitation to `set_task_metadata()` Cypher function, fixing injection vulnerability
- `Task.command` property now appends values of checkpoint requirement `Hint` fields `restart_parameters` and `bee_checkpoint_file__` to end of command list if they exist
- Removes the "unique task name" constraint in Neo4j
- Add `test_restart_task` to `test_wf_interface.py`, fix broken metadata

### DIFF
--- a/src/beeflow/common/gdb/neo4j_cypher.py
+++ b/src/beeflow/common/gdb/neo4j_cypher.py
@@ -1,5 +1,7 @@
 """Neo4j/Cypher transaction functions used by the Neo4jDriver class."""
 
+from re import fullmatch
+
 
 def constrain_tasks_unique(tx):
     """Constrain tasks to have unique names."""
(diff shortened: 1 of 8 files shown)


src/beeflow/common/gdb/neo4j_cypher.py
```diff
@@ -399,16 +421,12 @@ def set_task_metadata(tx, task, metadata):
...
     """
-    metadata_query = "MATCH (m:Metadata)-[:DESCRIBES]->(:Task {id: $task_id})"
-
-    # TODO: remove injection vulnerability before release
     for k, v in metadata.items():
-        if isinstance(v, str):
-            metadata_query += f" SET m.{k} = '{v}'"
-        else:
-            metadata_query += f" SET m.{k} = {v}"
+        if fullmatch(r"[A-Za-z_][0-9A-Za-z_]*", k) is None:
```
rstyd: Could use a small comment here just saying that this code is to prevent an injection vulnerability.

src/beeflow/common/gdb/neo4j_driver.py
```diff
@@ -126,6 +126,26 @@ def initialize_ready_tasks(self):
...
+        :type new_task: Task
+        """
+        with self._driver.session() as session:
+            session.write_transaction(tx.create_task, task=new_task)
+            session.write_transaction(tx.create_task_hint_nodes, task=new_task)
+            session.write_transaction(tx.create_task_requirement_nodes, task=new_task)
+            session.write_transaction(tx.create_task_input_nodes, task=new_task)
+            session.write_transaction(tx.create_task_output_nodes, task=new_task)
+            session.write_transaction(tx.create_task_metadata_node, task=new_task)
+            session.write_transaction(tx.add_dependencies, task=new_task, restarted_task=True)
```
rstyd: Is this call to add_dependencies missing the old_task?

src/beeflow/common/gdb/neo4j_cypher.py
Boogie3D: Added this comment

src/beeflow/common/gdb/neo4j_driver.py
Boogie3D: Add missing old_task parameter

rstyd:
LGTM!

Issue #450, opened 2022-05-12
Checkpoint/Restart Milestone 


rstyd:
This issue will track the work on our checkpoint/restart milestone. 

- [x] #451 
- [x] #452 
- [x] #453
- [x] #454 
- [x]  #455

Modify CWL Code to support a checkpoint requirement  
- 
- Will be called beeflow:CheckpointRequirement
- Will have the following parameters 
    - enabled: \<`true` or `false`\>
    - checkpoint_path: \<Path to directory containing checkpoint files\>
    - checkpoint_regex: \<Regex for the checkpoint files\>
    - restart_paramaters: \<parameter used for restarting the application with the checkpoint file\>
    - num_restarts: \<The number of restarts to attempt\> (Could be replaced with some sort of progress metric)
     
Create New Checkpointing CLAMR Workflow
-
- Will be like the old CLAMR workflow but with checkpointing enabled
- Can enable checkpointing with the -c n parameter which says write a checkpoint to disk every n time steps 
- beeflow:CheckpointRequirment will be
    - enabled: true
    - file_path: checkpoint_output
    - file_regex: backup[0-9]*.crx<sup>1</sup> 
    - restart_parameters: -R $(checkpoint_file) <sup>2</sup> 
    - num_tries: 3

Modifications to WF Interface / GDB Driver 
-
- Add a new task state `RESTARTED` that indicates a task has been restarted using checkpoint/restart
- Add a new function to the WF interface and GDB driver called `restart_task()` that creates a new task that's a child of the original.
    - restart_task will take the task object along with a path to the most recent checkpoint file. 
    - The original task will be marked as `RESTARTED` and the restart task will be marked as `READY`.
    - The restart task will be the original task with the restart parameters and checkpoint file inserted into the command line options

Modifications to WF Manager
-
- If a checkpointed task fails, attempt to restart the task with the `restart_task()` function. 
    - WFM will find the most recent checkpoint file in the `workflow_workdir`<sup>3</sup> and call restart_task with that checkpoint file
        - If no checkpoint files are found, the workflow fails with `UNRECOVERABLE_FAILURE`
    - Additional restarts will be done for this task up to `num_tries` times where `num_tries` is the number of attempts allowed in the CWL file. 
- After the checkpoint task has been created, it'll be submitted to the task manager and behave just like any other task. 

Modifications to Task Manager / Worker
-
- When a checkpoint task fails, the task manager will give a special message to the WFM when it updates the state
[...]

PR #468, opened 2022-06-06
Checkpoint restart milestone

pagrubel:
This PR will satisfy the milestone for adding the capability Checkpoint and Restart to workflows and addresses issue #450


pagrubel (commented):
We have some issues, like supporting LSF on summit and making checkpointing more flexible. I think those can be separate PR's. Please approve if you agree @Boogie3D @rstyd @jtronge

jtronge:
I noticed that the "Workflow completed" message isn't getting printed by the WFM with a checkpoint workflow.

pagrubel:
Checkpointing works for the clamr-checkpointing examples on fog and fails when the number of tries are exceeded. Also the cannon clamr example works with the most recent revisions. The WF manager updates the status of the workflow in the log. This PR is ready to be merged @rstyd

PR #471, opened 2022-06-09
Jinja Fix Issue #460

jtronge:
This should fix issue #460. I added a new ContainerRuntimeResult class which now includes an `env_code` parameter. I tested this on Darwin.


rstyd (approved):
LGTM

PR #474, opened 2022-06-15
Fix workflow_completed() not accounting for restarted tasks

Boogie3D:
- Fixed bug wherein `WorkflowInterface.workflow_completed()` was checking if **all** task nodes have state `COMPLETED` to determine if a workflow had completed.
  - This did not account for restarted tasks which have state `RESTARTED`
- This method now checks if only the final tasks in the workflow have state `COMPLETED`

### DIFF
--- a/src/beeflow/common/gdb/gdb_driver.py
+++ b/src/beeflow/common/gdb/gdb_driver.py
@@ -260,7 +260,7 @@ class GraphDatabaseDriver(ABC):
     def workflow_completed(self):
         """Determine if a workflow has completed.
 
-        A workflow has completed if each of its tasks has state 'COMPLETED'.
+        A workflow has completed if each of its final tasks has state 'COMPLETED'.
 
         :rtype: bool
         """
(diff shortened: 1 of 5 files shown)


rstyd (approved):
LGTM!

pagrubel:
Should the message be in the log now or does this need to be merged for that to work? I ran the checkpoint clamr example with this branch but the status of the worklfow is Running even though it actually made the movie.
@rstyd @Boogie3D

pagrubel:
There still is not a message in the log and the wf status is RUNNING although all tasks completed

Boogie3D:
Funny, it passed my unit tests. Looking into it.

pagrubel:
Just for clarity, I ran a the clamr-wf-checkpoint example with a timelimit in the jinja file. The workflow ran clamr, clamr(1) and ffmpeg to completion
Here is the result of a query:
4
What is the workflow id?
$ 39bf2e
clamr--RESTARTED
ffmpeg--COMPLETED
clamr(1)--COMPLETED

Boogie3D:
@pagrubel Think I got it, try now.

jtronge (approved):
This worked for me on Darwin.

pagrubel (approved):
Both clamr with and with checkpointing examples work now!

Issue #470, opened 2022-06-09
Show copyContainer is a BEE extension to DockerRequirements


pagrubel:
Rename copyContainer to BEE:copyContainer

pagrubel:
To be consistent with the MPI Requirements it should be beeflow:copyContainer

jtronge:
We may want to look into using a beeflow-specific container requirement. Something such as `beeflow:ContainerRequirement` that includes `copyContainer` and other container options that we use.

pagrubel:
This has been satisfied with PR#480

PR #480, opened 2022-07-05
Add beeflow:copyContainer to container driver and useContainer example

jtronge:
This should address issue #470. I'm marking this as WIP right now since I need to do a couple more tests.

Also I realized that I was wrong about where we can add CWL extensions. Looking further at [Arvado's extensions](https://doc.arvados.org/v2.4/user/cwl/cwl-extensions.html), I see that they use a `arv:dockerCollectionPDH` underneath `DockerRequirement`.

I also noticed that other CWL engines use a `$namespaces` parameter within the CWL file to indicate extensions namespaces. We probably want to look at this a little more because it may affect how compatible our CWL is with other implementations. I may make an issue for this.

### DIFF
--- a/src/beeflow/wf_manager.py
+++ b/src/beeflow/wf_manager.py
@@ -666,7 +666,7 @@ class JobUpdate(Resource):
                         submit_tasks_tm(tasks)
 
         elif job_state == "FAILED" or job_state == "TIMEOUT":
-            if 'task_info' in data:
+            if 'task_info' in data and data['task_info'] is not None:
                 task_info = jsonpickle.decode(data['task_info'])
                 checkpoint_file = task_info['checkpoint_file']
                 new_task = wfi.restart_task(task, checkpoint_file)
--- a/src/beeflow/common/build/build_driver.py
+++ b/src/beeflow/common/build/build_driver.py
@@ -134,7 +134,7 @@ class BuildDriver(ABC):
                     (self.dockerLoad, 'dockerLoad', 6, True),
                     (self.dockerFile, 'dockerFile', 7, True),
                     (self.dockerImport, 'dockerImport', 4, True),
-                    (self.copyContainer, 'copyContainer', 3, True),
+                    (self.copyContainer, 'beeflow:copyContainer', 3, True),
                     (self.dockerImageId, 'dockerImageId', 1, False),
                     (self.containerName, 'containerName', 2, False),
                     (self.dockerOutputDirectory, 'dockerOutputDirectory', 0, False)]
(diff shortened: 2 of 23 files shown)


jtronge:
I think this should be ready to go. All the examples should have the `beeflow:` prefix now, and I've added a new example for the `beeflow:useContainer` option.

src/beeflow/wf_manager.py
```diff
@@ -666,7 +666,7 @@ def put(self):
                         submit_tasks_tm(tasks)
 
         elif job_state == "FAILED" or job_state == "TIMEOUT":
-            if 'task_info' in data:
+            if 'task_info' in data and data['task_info'] is not None:
```
pagrubel, line 669: Remind me what is task_info do we really need to check it if the job failed or timed out

pagrubel: @rstyd can you tell me if we should check for task_info if a job_state is FAILED or TIMEOUT?

pagrubel:
I'm wondering if we should have a check for the container in useContainer (maybe copyContainer also) in the builder pre processing before submitting a job and waiting for an error?
This is from the error output of the job when I purposely put in the wrong container (doesn't exist).
```
input:   tar       /usr/projects/beedev/clamr/clamr-tos.tar.gz
srun: error: fg006: task 0: Exited with exit code 1
output:  dir       /tmp/clamr-tos.tar
error: not found: /usr/projects/beedev/clamr/clamr-tos.tar.gz
ch-run[16332]: error: can't stat: /tmp/clamr-tos.tar: No such file or directory (ch_core.c:245 2)
srun: error: fg006: task 0: Exited with exit code 1
```

I verified the same happens with copyContainer

src/beeflow/wf_manager.py
jtronge: I'm not sure exactly what's getting passed over in task_info. My jobs failed a couple times while testing and I noticed that this section of code was throwing an error. Maybe this check can be removed, but I wasn't sure.

jtronge:
Yeah I think some sort of validation will need to be done here. Perhaps it should be done with some sort of builder preprocessing step. I think we'd need to also have some other method of alerting the wfm when this kind of error happens for a task.

src/beeflow/wf_manager.py
rstyd: Hey so the way this works is task_info contains the checkpoint information if the task fails or times out. 

https://github.com/lanl/BEE_Private/blob/540fbeea1807b9126a79ad35d4bd56bb524a802f/src/beeflow/task_manager.py#L225-L231

pagrubel (approved):
All looks Good!

src/beeflow/wf_manager.py
jtronge: Ok, thanks. I added the `data['task_info'] is not None` check there because it was throwing something about the value being None when the wfm tried to use the it. I wasn't sure whether this is a problem with the WFM or with the TM. This was for when a task failed.

I tried finding the exact error in my logs, but I must have deleted it.

PR #481, opened 2022-07-06
Documentation

pagrubel:


pagrubel:
See README.rst in doc for building documentation.

pagrubel:
I think this first version of documentation is ready to add. Please review it. I'd like to get it in so we can start the publishing process.

Boogie3D (changes requested):
> Creating Configuration File:
>
> You will need to setup the bee configuration file that will be located in:
>
>    $HOME/.config/beeflow/bee.conf

This is fine but just keep in mind it is currently [bugged](https://github.com/lanl/BEE_Private/issues/499) on MacOS and Windows systems and will be fixed eventually.

> **Todo: Add cautions about which values should probably remain the same**

You may want to satisfy this Todo in Installation Guide and remove it.

The Installation Guide should probably also explain that client and backend may be separate installations.

The CLAMR build workflow figure has red squiggles under dockerFile and containerName. Bit of a nitpick but maybe try to get rid of those if it isn't a hassle.

Can we get a PNG of the new BEE logo with a transparent background to see what it might look like on the blue background?
image

pagrubel:
I think we need to get this in. I would like to work on most of those once I can see how we actually install BEE when it is packaged. 
The transparent one did not work well on the blue back ground since it is grey.

pagrubel:
Changing label to WIP so CI won't run for documentation only changes, however I'd still like to see this PR merged so we have something to publish.

pagrubel:
I'm adding the png file so you can experiment with it.

Boogie3D (approved):

PR #484, opened 2022-07-12
Slurm worker tests

jtronge:
I've updated the slurm worker tests. These will require a valid bee.conf and can be run with `pytest src/beeflow/tests/test_slurm_worker.py`.

I'm marking this as WIP for now since there is one test that is failing on Darwin for some reason. I also need to add some test cases for when slurmrestd fails to start.


jtronge:
I figured out the issue with my tests. They were relying on temporary files in the `/tmp` directory which is not shared across nodes. Unfortunately this means that the slurm worker test will now create temporary files/dirs in our home dir.

src/beeflow/tests/test_slurm_worker.py
```diff
@@ -63,10 +63,10 @@ def decorator():
 
 
 def setup_worker_iface(fn):
-    """Add a deacorator that creates the worker interface, but not slurmrestd."""
+    """Add a deacorator that creates the worker interface but not slurmrestd."""
```
pagrubel:
```suggestion
    """Add a decorator that creates the worker interface but not slurmrestd."""
```

pagrubel (commented):
Minor spelling comment

src/beeflow/tests/test_slurm_worker.py
jtronge: Ah, good catch

pagrubel (approved):

Issue #419, opened 2022-02-04
Set Up CI


pagrubel:
Get ready for CI when the repo goes public
Start gathering requirements

jtronge:
I think we will probably need a couple different CI workflows to run:

* Some sort of integration test, that runs complete workflows with BEE (such as the CLAMR example) and then records the results
* Unit tests/general testing matrix - I'm guessing we would simply install BEE here and then run pytest

We may need some more, but I think these two are a good start. Both of these should be able to work with GitHub Actions, if that is the route we want to go.

Also, we should decide when these are run. For instance, do we want these to run when a developer opens a PR into develop? Do we want them to run on every merge as well? There are other possibilities as well.

pagrubel:
@jtronge would you like to add what else needs to be done for this or should we close this issue

jtronge:
I'll go ahead and close this, if that's ok. We'll probably need to make some more updates to the CI, but those are probably better off with new issues.

PR #489, opened 2022-07-15
Fix WF State

rstyd:
This PR fixes an issue where the workflow state wasn't getting set correctly after a workflow completed which was inhibiting #419. It also adds a new function in the wf_interface which can set a workflow's state to completed without finalizing it. Finally, this PR also fixes an issue in the crt_driver that was causing the load module step before a workflow to fail.


pagrubel (approved):
Looks good

jtronge (approved):
I tested this on my setup and it works.

PR #490, opened 2022-07-19
Fix pylama complaints for builder

pagrubel:


pagrubel:
A few things still need to be done:
      build_main in build_interfaces.py is too complex
      build/container_drivers.py also has some methods that are too complex
      containerName is a BEE extension so should be beeflow:containerName
      Build messages are too verbose, this could be another issue though

Boogie3D (approved):

PR #495, opened 2022-07-26
Change containerName to beeflow:containerName

jtronge:
This adds a prefix for the `beeflow:containerName` option which I forgot about in the previous PR. Right now the examples for this are both single-file noyaml examples which currently don't work with the bee_client. This can be tested directly through the builder. I may update one of the examples though to test the bee_client.


pagrubel:
I'll put an example in the usual place for you to work with

PR #502, opened 2022-08-01
Issue433/cleanup modules

pagrubel:
@Boogie3D I added the cleaner task_manager, it now only has two ignores.


Boogie3D (approved):
Everything checks out but wf_manager.py which is currently being refactored.

Issue #433, opened 2022-02-16
Clean up Modules:


pagrubel:
Fix messy code. Remove todos and such from code. Make sure all code passes pylama checks.

- [x] Graph database - Steven
- [x] Parser - Steven
- [ ] Workflow Manager - Rusty
- [ ] bee_client - Rusty
- [x] Task Manager - Pat
- [x] Container Runtime - Pat
- [x] Scheduler - Jake
- [x] Build - Pat
- [x] Cloud - Jake
- [ ] cli.py renamed to beeflow.py (Do we want to do this?)
- [x] Config including bee_cfg - Jake
- [x] logging
- [ ] tests

Please add to the above or reassign.

pagrubel:
Moved this to the active column. Steven has started the branch
https://github.com/lanl/BEE_Private/tree/issue433/cleanup_modules

pagrubel:
It's has been decided to work in your own branch the name should probably be issue433/____________  and make the PR to merge into develop

Issue #333, opened 2021-05-21
Scheduler Improvements


jtronge:
I'm creating this issue to keep track of a number of improvements needed for the scheduler:

* The main jobs endpoint needs to be able to receive updates from the WFM when jobs are updated
* The scheduling endpoint should return a simple mapping from job_name -> resource_id, rather than a complicated task object
* Scheduler should keep track of what resources are currently being used
* Resources should have bandwidth information
* Cost metrics need to be implemented (i.e. cost per resource usage, cost per task)
* ML part of the scheduler could be reworked to focus on task runtime estimation - this may make more sense for scheduling purposes

jtronge:
As I start working on #433 I've noticed a couple additional cleanup items that I need to fix sometime soon:

- Logging needs to follow the conventions set in the other modules
- Configuration needs to simplified (it is extremely complex right now)
- Should we keep the MARS scheduler?

pagrubel:
Do you think the MARS scheduler will be helpful going forward?

jtronge:
I think we would need a use case for it and more training data. Right now we don't have enough real-world training data to actually ensure that it's doing what it should be doing. The scheduler is kind of designed for running with multiple task managers on different systems and I think we need to make some more design decisions before we integrate that code in.

PR #512, opened 2022-08-05
Scheduler Cleanup for Issue #433

jtronge:
This should address most of issue #433. I'm also thinking about removing some of the modules that aren't really useful as is, such as the MARS component.

I would also like to try to do a refactor of the scheduler in a future PR. Scheduler improvements are documented in #333.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -15,14 +15,16 @@ ignore = R0902,R0903,R0904,R0912,R0913,R0914,R0916,W0603,W1203,C0413
 
 [pylama:pylint]
 max_line_length = 99
-good_names = i,j,k,v,t,m,s,n,ex,fp,id,tx,fn,Run,_
+good_names = i,j,k,v,t,m,s,n,x,y,ax,ex,fp,id,tx,fn,pi,Run,_
 
 [pylama:pycodestyle]
 max_line_length = 99
-ignore = E402
+ignore = E402,W503
 # E402: module level imports should be at the top of the file
+# W503: This warns about a line break before a binary operator. I've noticed this
+#       conflicting with the line length warning several times, especially with
+#       code using list comprehensions.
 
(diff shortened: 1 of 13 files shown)


pagrubel:
If this is ready for review please remove the WIP label

setup.cfg
```diff
@@ -15,14 +15,14 @@ ignore = R0902,R0903,R0904,R0912,R0913,R0914,R0916,W0603,W1203,C0413
...
 max_line_length = 99
-good_names = i,j,k,v,t,m,s,n,ex,fp,id,tx,fn,Run,_
+good_names = i,j,k,v,t,m,s,n,x,y,ax,ex,fp,id,tx,fn,pi,Run,_
 
 [pylama:pycodestyle]
 max_line_length = 99
-ignore = E402
+ignore = E402,W503
 # E402: module level imports should be at the top of the file
+# W503: this frequently conflicts with the line length warning
```
pagrubel: What is the warning "this"

src/beeflow/scheduler/train.py
```diff
@@ -123,3 +122,12 @@
     actor.predict(vec)
     critic.predict(vec)
     mars.save_models(actor, critic, args.trained_model)
+
+
+if __name__ == '__main__':
+    main()
+# Ignore R0915: This requires a refactor related to issue #333.
```
pagrubel: Again say what you are ignoring

pagrubel:
If we are going to ignore pylama explain what you are ignoring so someone reviewing the code knows what it is, otherwise all looks good

jtronge:
Ok thanks. I think I've updated most of these. I hope to remove a lot of these ignores when I get time to work on issue #333.

pagrubel (approved):

PR #516, opened 2022-08-09
Use `bee_cfg show` to display the contents of the current bee.conf

jtronge:
This changes `bee_cfg list` to `bee_cfg show`.


pagrubel:
Adding documentation to this PR so also adding the no ci label

pagrubel (approved):

PR #525, opened 2022-08-15
Set version for pre-release; add other logos to try out; fix clamr-st…

pagrubel:
…ep image


pagrubel:
@Boogie3D please review this so we can make our repo public with a pre-release version towards packaging

Boogie3D (approved):

PR #526, opened 2022-08-16
Add Build Integration Tests and Remove CLAMR

jtronge:
This removes the CLAMR workflow from the integration tests, and adds in a number of different tests for the builder. This should test `beeflow:copyContainer`, `beeflow:useContainer`, `dockerFile` and `dockerPull`. Instead of relying on workflows in the `src/beeflow/data/` or `examples/` directories, the workflows and Dockerfiles are generated at test time from the code itself.

### DIFF
--- a/ci/README.md
+++ b/ci/README.md
@@ -1,0 +1,25 @@
+# CI code
+
+This directory contains all the scripts that are needed for configuring and
+running BEE on a CI machine. The scripts here are as follows:
+
+* `env.sh`: CI environment set up
+* `bee_install.sh`: script for installing BEE
+* `bee_start.sh`: start script for BEE
+* `deps_install.sh`: BEE external dependency install script (this installs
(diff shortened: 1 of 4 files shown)


jtronge:
Also, you should be able to run this test locally (I ran it on darwin). After starting bee with `beeflow`, you should be able to run `./ci/integration_test.py`.

pagrubel:
Can you add that instruction in a Readme in the ci directory, if you haven't already

jtronge:
Ok I added the README.

pagrubel:
Probably should not use ch-image reset when running on a cluster. You could change it to ch-image delete and delete the images you created in the test

pagrubel (changes requested):
Change ch-reset to ch-delete (specific images)

pagrubel (approved):

pagrubel (approved):

PR #527, opened 2022-08-16
Satisfy pylama for log and some misc. files

pagrubel:
I had to ignore moving the import bee stuff; if one of you can change that I'm all for it.


src/beeflow/tests/test_tm.py
```diff
@@ -73,22 +73,22 @@ def test_remove_task(flask_client, mocker):  # noqa
     """Test cancelling a workflow and removing tasks."""
     # Add a few tasks
     beeflow.task_manager.job_queue.append({2: {'name': 'task1',
-                                                            'job_id': 1,
```
jtronge, line 76: It complains about the order of the imports in this file.

jtronge (commented):
It has a few warnings in `src/beeflow/tests/test_tm.py`. When I ran `pylama src/beeflow/common/worker/` it also gave some warnings for `src/beeflow/common/worker/worker.py`, but those can probably be done in a different PR.

src/beeflow/tests/test_tm.py
pagrubel: I thought I fixed those... I'll look again

pagrubel: Latest commit should fix those

jtronge (approved):

Boogie3D (approved):

Issue #211, opened 2020-08-13
Refactor Workflow Manager


rstyd:
Refactor the workflow manager to remove all the lama issues and consider Quincy's notes.

PR #532, opened 2022-08-23
WFM Refactor

rstyd:
This PR addresses #211. It breaks up the currently monolithic workflow manager code into multiple resource sections that each operate a particular REST endpoint. Each endpoint is small self contained script which are then called by the `wf_manager.py` script which actually starts flask. The StartGDB script has been reworked into a dependency manager that will support running any dependency through charliecloud or singularity. (Currently it just has charliecloud support) The dependency manager script also intelligently builds the BEE dependency container so we don't have to repetitively build container images. Additionally, this PR adds support for a workflow manager database in SQLite that aids managing multiple workflows. Finally, it comes with a new set of workflow manager unit tests with pytest mockups. Currently, the tests are in the wf_manager directory just for final testing. They'll be moved over shortly.


pagrubel:
What is the status of this PR?

jtronge:
I'm getting a `[01;31mERROR: Workflow Manager failed to start. Exiting.[0m` from beeflow when I try to start it. The `wf_manager.log` also isn't generated, so I'm guessing it's a problem in cli.py.

pagrubel:
What happened to the CI tests?

jtronge (approved):
Both the clamr-ffmpeg and the cat-grep-tar workflows run fine for me on this branch. I ran `pylama beeflow/wf_manager` from the root and it gave no errors/warnings. I'm approving this.

pagrubel (approved):
Workflows work fine. Two small issues I see is having workflows.db in the place that beeflow is started and the dependency containers unpacked in the workdir instead of memory. I'll make issues for these since they can be fixed in another PR and will be for the multiple workflow PR.

PR #534, opened 2022-08-25
Small CWL Example and Stdout Handling

jtronge:
This adds a simple CWL example that could be useful for new users and also attempts to add some form of stdout handling.

I've marked this as a WIP because the current stdout implementation won't work for containers and it also just dumps files in whichever directory the TM is started in. @Boogie3D Do you think we would be able to attach some sort of `output_dir` to a workflow? I'm wondering if we would be able to use this in the `task.command` method to give files an absolute path so that we don't have to worry about the working directory of the job.

The new example has one curl task which downloads a file, then two grep tasks that check for different key words in the file and then a final tar task which tars the grep output files up. You should be able to create the workflow tarball with `tar -C examples/ -cf simple-workflow.tgz curl-grep-tar` and then run the example with `bee_client submit simple-workflow simple-workflow.tgz workflow.cwl input.yml`.


beeflow/common/worker/worker.py
```diff
@@ -82,19 +82,24 @@ def prepare(self, task):
...
-        # workflow_path = f'{self.workdir}/workflows/{task.workflow_id}/{task.name}-{task.id}'
         task_save_path = self.task_save_path(task)
         crt_res = self.crt.run_text(task)
         requirements = dict(task.requirements)
         hints = dict(task.hints)
+        main_command = crt_res.main_command[:]
+        # Append a tee to a file if requested
+        if task.stdout:
+            main_command.extend(['|', 'tee', task.stdout])
+        # TODO: It seems to me that there might be a better way to save stdout,
```
pagrubel, line 93: I think we shouldn't have todo's in the source file, instead use issues. You can quote the code lines there.

pagrubel (changes requested):

beeflow/common/worker/worker.py
jtronge: Hmmn, I think it's a good idea to have both issues and TODOs in the code. Perhaps we could reference the issue in the comment (something like `TODO:123: ...` for issue 123). Quoting code lines in the issue will quickly get out of date as code is refactored or the lines change as new code is added to the file. The TODO acts a marker in the code indicating exactly where something needs to be fixed/reviewed.

jtronge:
This should be working now (both with and without containers). I'm going to add some documentation about the example and then remove the WIP.

jtronge:
Note that this includes a new `workdir` option for bee workflow submissions. This is where all input files are expected to be and where all commands will be run. For running containers, this requires that the container runtime append a `--cd $WORKDIR` to the `ch-run` command and as a result of this, we can't have `--cd` in `chrun_opts`.

Boogie3D (dismissed):

beeflow/common/config_driver.py
```diff
@@ -336,7 +336,19 @@ def bee_workdir_init(path, _cur_opts):
                   depends_on=('task_manager', 'container_runtime', 'Charliecloud'))
 VALIDATOR.option('charliecloud', 'image_mntdir', attrs={'default': join_path('/tmp', USER)},
                  info='Charliecloud mount directory', validator=validate_make_dir)
-VALIDATOR.option('charliecloud', 'chrun_opts', attrs={'default': f'--cd /home/{USER}'},
+
```
pagrubel, line 339: I'm still struggling with why would not allow a user to have --cd /home/... as an option for charliecloud. It seems we are cutting off an option they may want.

pagrubel (approved):

PR #535, opened 2022-08-25
Delete accessing src from unit tests

pagrubel:
This correct running the unit tests.


jtronge (approved):
Looks good. I'm not sure why the wfm unit tests are still failing. I think some of those unit tests need to either have beeflow running or the gdb running. I'll try to make a PR sometime today or next week to fix that, as well as merge the unit and integration tests into one action.

PR #536, opened 2022-08-29
Combine unit tests and integration tests into one github action

jtronge:
This should combine the unit and integration tests into one action so that we are not wasting CI minutes doing a poetry install.

I added a unit test script for running pytest and recording which tests fail. Note that if any of the unit tests fail, this will now short circuit the integration test and the action will just fail. Since a lot of our unit tests are failing right now for some reason I've added an `ignore` option to the unit test script for broken tests. I think a lot of these fails have to do with context, for instance the GDB not running in the background.


pagrubel:
So if someone messes up a unit test, but the rest of the system is not affected integration tests would not occur? I'm not sure that's a good idea.

PR #546, opened 2022-09-19
WSGI

jtronge:
This is based on Rusty's original wsgi_final branch and should be able to run bee with wsgi. Right now the WFM is broken, but it should work once I merge in the workflow refactor code.

Some of the other changes that have been made are:

* Sqlite3 DB for the Scheduler and TM
* Use of sockets for REST calls
* Refactor of `beeflow`/`cli.py`
    * Use gunicorn for starting components
    * Handling of log output
    * Daemon support


jtronge:
I think this should be working now. The `examples/cat-grep-tar` workflow runs and the integration test worked locally.

I ran into some issues with `examples/clamr-ffmpeg-build`. I think we might have had this problem before, but since it relies on a relative path (`beeflow/data/dockerfiles/Dockerfile.clamr-ffmpeg`), it fails when beeflow isn't running in the repo directory. I'm thinking it might be a good idea to require Dockerfiles to be in the workflow tarball.

jtronge:
To launch everything, you can run `beeflow start`. By default this will launch a daemon process that then starts up all the components. This should still write a log to `~/.beeflow/logs/beeflow.log`. If you want to run it in the foreground you can launch with `beeflow start -F`.

You should also be able to kill everything with `beeflow stop` as well (that is everything except the gdb).

pagrubel:
> I ran into some issues with `examples/clamr-ffmpeg-build`. I think we might have had this problem before, but since it relies on a relative path (`beeflow/data/dockerfiles/Dockerfile.clamr-ffmpeg`), it fails when beeflow isn't running in the repo directory. I'm thinking it might be a good idea to require Dockerfiles to be in the workflow tarball.

That is a good idea!

beeflow/tests/test_scheduler_rest.py
```diff
@@ -92,7 +95,7 @@ def test_schedule_job_no_resources(scheduler):
     :param scheduler: url returned by scheduler fixture function
```
pagrubel: Did the param type change

pagrubel (commented):
A few operational things I would like to see is a statement like when envoking beeflow start, just a statement like 
Starting Task Manager, Scheduler and Work Flow Manager.

I would also like to have a beeflow status showing what is running (could indicate health of the components, also could alert a user that they are on the wrong front end by showing nothing is running)

Finally I think there should be a warning from beeflow stop if workflows are running, atleast until we have some sort of resiliency built in.

beeflow/tests/test_scheduler_rest.py
jtronge: Yes I think I had to change that. I'll need to fix that text there.

pagrubel (approved):

PR #552, opened 2022-09-22
Generate WF ID pre-parsing and support CWL stderr

### LINKED ISSUE #540: Workflow IDs Need to be Generated Before Parse Time
Currently, workflow IDs are assigned after a workflow has been parsed. This presents several challenges for the workflow manager. First, it has to allocate a temporary space for the workflow since the parser needs the workflow files to be on disk. Then, it parses those files and moves them to a workflow directory using the workflow ID. Additionally, when we launch the GDB with a container runtime, the mount point has to be given for each workflow which includes the GDB data directory. Since it's a temp directory at first, the whole thing is super clunky and doesn't really support running multiple workflows since you don't have anyway to connect the directory in which it is running with the workflow ID since it is unknown. 

I propose modifying the workflow interface such that the workflow manager can request a workflow ID for a new workflow, do all the required housekeeping _then_ parse the workflow. Right now the `parse_workflow` function returns a workflow interface object. So, we'll need detach that from the parsing part.

Boogie3D:
One major refactor and one new feature are added by this PR, which resolves #540:
1. Workflow IDs must be generated before parsing and assigned to the corresponding workflow via the parser. This change was made to accommodate the need to create a workflow directory using the workflow ID before parsing.
2. Support for specifying capture of `stderr` to a file in the CWL parser and graph database, akin to `stdout` which is already supported.

Main changes:
- `generate_workflow_id()` is now a function in `beeflow.common.wf_data`
- The `Workflow` class initializer and `WorkflowInterface.initialize_workflow()` now take a mandatory argument `workflow_id`.
- Likewise, `CwlParser.parse_workflow()` also takes a mandatory argument `workflow_id`, which is likely a breaking change for other components.
- The `Task` class now takes an optional argument `stderr`
- The Neo4j `Task` node now stores an additional property `stderr`.
  - When querying the database, the value of the `stderr` property is assigned back to the recreated `Task` object.
- `beeflow.tests.test_wf_interface` and `beeflow.tests.test_parser` have been updated and are passing.

### DIFF
--- a/beeflow/client/bee_client.py
+++ b/beeflow/client/bee_client.py
@@ -128,7 +128,7 @@ def submit(wf_name: str = typer.Argument(..., help='The workflow name'),
            yaml: str = typer.Argument(..., help='filename of YAML file'),
            workdir: pathlib.Path = typer.Argument(
                ...,
-               help='working directory for workflow containing input + output files',
+               help='working directory for workflow containing input + output files'
            )):
     """
     Submit a new workflow
@@ -150,7 +150,7 @@ def submit(wf_name: str = typer.Argument(..., help='The workflow name'),
         'wf_filename': os.path.basename(wf_path).encode(),
         'main_cwl': main_cwl,
         'yaml': yaml,
-        'workdir': workdir,
+        'workdir': workdir
     }
     files = {
         'workflow_archive': wf_tarball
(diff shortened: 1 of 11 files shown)


pagrubel:
@rstyd Do we need this PR or have you merged the changes into the Multiple Workflows PR #571

Issue #551, opened 2022-09-20
Require dockerFile to be included in the workflow tarball


pagrubel:
Idea from Jake:
I ran into some issues with examples/clamr-ffmpeg-build. I think we might have had this problem before, but since it relies on a relative path (beeflow/data/dockerfiles/Dockerfile.clamr-ffmpeg), it fails when beeflow isn't running in the repo directory. I'm thinking it might be a good idea to require Dockerfiles to be in the workflow tarball.

Any downsize? Should we allow the user to specify one that is external but copy it before packaging the directory or giving an error if it doesn't exist?

pagrubel:
Also, wanted to mention this is also good for provenance.

pagrubel:
Implemented in #560

PR #560, opened 2022-10-05
Fix for issue #551 - Require dockerfile in tarball

jtronge:
I think this should fix issue #551. I also updated the integration tests to make them a little easier to work with. We might want to add some that test for multi-step workflows soon.


pagrubel:
Satisfies issue #551

pagrubel (approved):
Looks good

PR #562, opened 2022-10-07
Checkpoint restart fix

jtronge:
This adds back the checkpoint-restart code from an earlier version, and also ensures that the checkpoint path is relative to the workdir.

### DIFF
--- a/beeflow/common/wf_data.py
+++ b/beeflow/common/wf_data.py
@@ -2,6 +2,9 @@
 from collections import namedtuple
 from uuid import uuid4
 from copy import deepcopy
+import os
+
+from beeflow.common.container_path import convert_path
 
 # Workflow input parameter class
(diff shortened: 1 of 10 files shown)


pagrubel:
This works unless file_path isn't declared, but it only fails upon trying to restart. Shouldn't we check for file_regex & file_path when parsing. I guess this could be another issue but we should maybe at least check if they exist. Everything works well for our example in beeflow/data/cwl/bee_workflows/clamr-wf-checkpoint and when I deleted the file_path an exception was thrown.

pagrubel:
I'm going to approve this with the caveat that we should think about what we check for at parsing a checkpoint/restart requirement.

pagrubel (approved):
This PR gives back the ability to use checkpointing and restarting for workflows.

PR #563, opened 2022-10-14
Error handling

jtronge:
This is one attempt to make error messages a little easier to work with for the user.

I modified the parser so that whenever it has an error, it throws a `CwlParseError`. The wf_manager catches this, saves the error message and then passes it back to the client in a 400 error response.

Error messages are still a little cryptic, since it sometimes prints the random directory where we untarred the workflow, but I think we can worry about this later.

### DIFF
--- a/beeflow/client/bee_client.py
+++ b/beeflow/client/bee_client.py
@@ -8,9 +8,11 @@ import os
 import sys
 import logging
 import inspect
-import shutil
 import pathlib
+import shutil
 import subprocess
+import textwrap
+import time
 import jsonpickle
 import requests
 import typer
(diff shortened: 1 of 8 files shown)


beeflow/client/bee_client.py
```diff
@@ -173,6 +197,9 @@ def submit(wf_name: str = typer.Argument(..., help='The workflow name'),
         error_exit('Could not reach WF Manager.')
 
     if resp.status_code != requests.codes.created:  # pylint: disable=no-member
+        if resp.status_code == 400:
```
rstyd, line 200: I think I get the logic, but why did you pick 400 in particular here?

beeflow/wf_manager/resources/wf_list.py
```diff
@@ -25,24 +25,13 @@
...
     cwl_path = os.path.join(workflow_dir, main_cwl)
     if yaml_file is not None:
         yaml_path = os.path.join(workflow_dir, yaml_file)
-        try:
-            wfi = parser.parse_workflow(cwl_path, yaml_path)
-        except AttributeError:
-            log.error('Unable to parse')
-            resp = make_response(jsonify(msg=parse_msg, status='error'), 418)
-            return resp
+        wfi = parser.parse_workflow(cwl_path, yaml_path)
```
rstyd, line 32: This is a great change!

beeflow/common/api.py
```diff
@@ -0,0 +1,12 @@
+"""BeeApi wrapper around Flask-Restful for handling exceptions."""
+import traceback
+from flask import make_response, jsonify
+from flask_restful import Api
+
+
+class BeeApi(Api):
+    """Wrapper around Flask-Restful's API to catch exceptions."""
+
+    def handle_error(self, e):  # noqa (conflict on naming in base class vs. following convention)
```
rstyd, line 10: When does this function in the wrapper get called? It's not immediately obvious to me how an exception would trigger it.

jtronge: I actually had to dig around for this a little. Flask itself has an error handler/decorator function: https://flask.palletsprojects.com/en/2.2.x/errorhandling/#registering. I tried that first, but nothing was happening when a 500 occurred.

I found out that Flask-Restful is actually doing some extra error catching here, see [this](https://stackoverflow.com/questions/34037283/flask-errorhandler-does-not-work-in-gunicorn-after-introducing-flask-restful). Their documentation isn't very good in this area and they only seem to "officially" support attaching a specific error message like in that example. This seems to be the only to catch 500s here.

Of course this was kind of an experiment. Maybe we don't want to pass the 500 error details back to the client and can just leave it in the logs. I was thinking that ideally we don't want the 500 to happen for a user, and if it does we should make it easy for them to wrap up their stuff and send us something that we can use to diagnose the problem.

beeflow/client/bee_client.py
jtronge: I think 400 means something like the user submitted bad arguments or a bad request (not our fault). So this would be returned if the user mistyped something or sent a bad workflow file.

rstyd: Sounds reasonable.

beeflow/common/api.py
rstyd: Huh cool

rstyd:
Looks good to me!

rstyd (approved):

PR #566, opened 2022-10-19
529 version command

### LINKED ISSUE #529: beeflow --version command needed
Might be nice to support this version command

aquan9:
This resolves issue #529. I was able to get beeflow --version working (hopefully without breaking anything else).

### DIFF
--- a/beeflow/cli.py
+++ b/beeflow/cli.py
@@ -324,20 +324,22 @@ def stop():
     beeflow_log = log_fname('beeflow')
     print(f'Beeflow has stopped. Check the log at "{beeflow_log}".')
 
-@app.callback(invoke_without_command=True)
-def version(version : bool = False):
-    #Print out the current version of the app and then exit. 
-    if version:
-        #Get the path with the pyproject.toml file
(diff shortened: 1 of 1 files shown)


jtronge:
Ok, great. I tried this and it works. If we need to do anything more complicated then it might be worthwhile adding a toml parser, but I think this is good for now.

You'll need to fix the pylama lints and then I think it should be good to go (`pylama ${file}` will show what needs to be done).

pagrubel:
I'm not sure this will work if someone installs beeflow with pip.

aquan9:
> I'm not sure this will work if someone installs beeflow with pip.

@pagrubel Does a pip install change the BEE directory structure somehow? Which part of the code are you concerned with?

pagrubel:
I don't know if the files are actually available when we do a pip install.

pagrubel:
I guess it may be time to publish again and try it out

pagrubel (approved):
Okay you can ignore my previous comments I think it will work with pip.

PR #570, opened 2022-10-21
Add a subsection for Slurm job output

jtronge:
This adds details on how to find job output for tasks of a workflow.

### DIFF
--- a/docs/sphinx/advanced_usage.rst
+++ b/docs/sphinx/advanced_usage.rst
@@ -29,11 +29,12 @@ script will go to files in
     The task/step name used within the workflow submission
 
 Within this directory, you can find the submission script and two output files
-for stderr (extensions ``.err``) and stdout (extension ``.out``).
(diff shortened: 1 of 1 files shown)


pagrubel (changes requested):
This output is also saved in the archive of a workflow for provenance so that should be mentioned. (That is also a good reason not to change that part of the jinja template, but that documentation is in another place.)

pagrubel:
I'm wondering if we want people to examine the errors themselves first.
So I'm thinking this should be changed: 
"If something bad happens, such as if BEE fails unexpectedly, then you may want to wrap up these output files as well as BEE’s general logs and send them to a BEE developer for further inspection."

Something about them looking at the errors in the err file of the task that fails If they need help contact us and we may direct them to do what you say above. 

When I add the documentation about the logs I may direct them to that in this space also if it seem appropriate.

pagrubel (approved):
Looks good, thank you!

Issue #271, opened 2021-01-11
BEE Multi-Workflow Support


rstyd:
BEE needs to support running multiple workflows for a single user across various computing resources. At the moment, BEE only works for a single workflow. Below, I’ll summarize what changes are required to support multiple workflows for each component. I did not consider running across clusters since that will complicate things dramatically. 

Workflow Manager:
          The largest change required for the WFM is keeping track of multiple GDBs. We can do this by adding a dictionary of workflow_interface objects that are mapped to a workflow ID. The WFM already requires a workflow_id for most REST endpoints, but we’re not currently using it. With the inclusion of the dictionary, each endpoint function can just work on the specified workflow. We will need to add the workflow_id to the JobUpdate endpoint for updating task states. 

GDB: 
         We need a workflow_id to identify each distinct workflow. Currently, this information is hardcoded when workflows are created. From previous discussions, the workflow_id should be stored directly in the GDB. This will require creating a new parameter in the GDB. 

Task Manager:
	The task manager generally doesn’t care which Workflow Manager it receives jobs from. It just runs them on slurm/lsf. The one thing that needs to be changed is when the TM updates the WFM with the task state. Currently, this is done with a task_id through the JobUpdate endpoint on the WFM. We’ll update the WFM to use the workflow_id as well as the task_id for the JobUpdate endpoint. This does pose one problem. The TM currently only receives batches of task objects which do not have workflow_ids attached to them. We’ll need to either add workflow_ids to these task objects or come up with another way to bind this information to each task. 

Client:
	The client fully supports multiple workflows. 

Scheduler:
	As currently defined, as long as the jobs the scheduler receives are independent, it should be able to schedule them with no issue even if they are from different workflows. The one caveat is this requires the workflows to all use the same pool of resources/QoS on the same cluster.

rstyd:
This issue has been resolved by #571.

PR #571, opened 2022-10-21
Multiple workflows

rstyd:
This PR addresses #271. There were a ton of conflicts so currently making sure everything works as it should.


pagrubel:
Since this is  a large PR I'd like to see 2 reviews before merging, so I'm assigning Jake and myself.

pagrubel:
Deleted that last comment, but I do think we should have 2 approvals before merging this.

jtronge:
Some of the unit tests are failing because of changes to the gdb start/stop code. I wonder if there's some way we can make it easier to start the gdb for testing.

jtronge:
I get this when trying to run the cat-grep-tar workflow: `AttributeError: module 'beeflow.wf_manager.resources.wf_utils' has no attribute 'create_workflow_dir'`

jtronge:
Ok, I tried submitting two clamr workflows at the same time to different workdirs and they both worked.

Do you think we could add the workflow id (or the short version) in the log output. Something like this maybe:

```
INFO: Submitted ['clamr'] to Task Manager (wf_id: a55624)
INFO: Workflow Completed (wf_id: a55624)
```

pagrubel:
Show me that stderr is working I cannot get output from ffmpeg other than our BEE output in the .beeflow/workflows dir

beeflow/tests/gdb.py
```diff
@@ -1,21 +1,26 @@
...
 import time
 from beeflow.wf_manager.common import dep_manager
+from beeflow.wf_manager.resources import wf_utils
 
 
 def start():
     """Start the GDB."""
-    dep_manager.kill_gdb()
-    dep_manager.remove_current_run()
+    # dep_manager.kill_gdb()
```
pagrubel, line 9: Why are you keeping this around?

beeflow/tests/gdb.py
```diff
@@ -1,21 +1,26 @@
...
+    # dep_manager.remove_current_run()
```
pagrubel, line 10: Is this dangerous to keep this around?

pagrubel:
> Show me that stderr is working I cannot get output from ffmpeg other than our BEE output in the .beeflow/workflows dir

Okay we decided that we should make sure stderr is fully supported for the release but it should be fixed in another PR

jtronge (approved):

pagrubel (approved):

Issue #294, opened 2021-02-26
logging: Add timestamp to log formatting


qwofford:
We generate a lot of log files. Since they are logging asynchronously, timestamps would be helpful. Add a timestamp to the logging tool.

PR #573, opened 2022-10-27
Add formatting improvements to the logger and remove unused logs

jtronge:
This removes some of the existing log code and changes the log format to be '[timestamp] module_path:function() message'. The code for creating extra logs, such as in the container driver or slurm worker, has been removed since it was being duplicated in logs for the main components.

Also, if you want to use a log with this, you'll need to call the `setup()` function, something like this:

```
log = bee_logging.setup(__name__)
```

### DIFF
--- a/beeflow/common/build/container_drivers.py
+++ b/beeflow/common/build/container_drivers.py
@@ -8,10 +8,12 @@ import shutil
 import subprocess
 import tempfile
 from beeflow.common.config_driver import BeeConfig as bc
-from beeflow.common.log import main_log as log
+from beeflow.common import log as bee_logging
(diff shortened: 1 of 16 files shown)


jtronge:
This addresses #294.

pagrubel:
This looks great to me. So much more information and those extra duplicate logs are gone. Bonus, we now  even have build information (i.e. output when building a container in the log) It makes it longer but at least if a container is expected to be built and it fails during the build, we will have information.

pagrubel (approved):

rstyd (approved):
Looks good to me!

PR #585, opened 2022-11-06
475 allow workflows from directories

### LINKED ISSUE #475: Allow workflows to be selected from directories as well as tar balls
For ease of use don't require the user to tar up a directory for the cwl. Allow them to use a directory with the cwl files.

aquan9:
This PR allows you to do a bee_client submit and it will automatically do the packaging step if the workflow path is found to be a directory.

### DIFF
--- a/beeflow/client/bee_client.py
+++ b/beeflow/client/bee_client.py
@@ -173,6 +173,7 @@ def submit(wf_name: str = typer.Argument(..., help='The workflow name'),
            workdir: pathlib.Path = typer.Argument(...,
            help='working directory for workflow containing input + output files',)):
     """Submit a new workflow."""
+    tarball_path = ""
     if os.path.exists(wf_path):
         # Check to see if the wf_path is a tarball or a directory. Run package() if directory
         if os.path.isdir(wf_path):
@@ -224,7 +225,8 @@ def submit(wf_name: str = typer.Argument(..., help='The workflow name'),
     logging.info('Sumit workflow:  {resp.text}')
 
     # Cleanup code
-    os.remove(tarball_path)
+    if tarball_path:
+        os.remove(tarball_path)
     return wf_id
 
 
--- a/beeflow/client/bee_client.py
+++ b/beeflow/client/bee_client.py
@@ -177,8 +177,11 @@ def submit(wf_name: str = typer.Argument(..., help='The workflow name'),
         # Check to see if the wf_path is a tarball or a directory. Run package() if directory
         if os.path.isdir(wf_path):
             print("Detected directory instead of packaged workflow. Packaging Directory...")
-            package(wf_path, pathlib.Path("/tmp"))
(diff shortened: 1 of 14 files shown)


aquan9:
My plan is to include the documentation updates in a separate PR since this one has such an old base.

beeflow/client/bee_client.py
```diff
@@ -167,14 +167,20 @@ def match_short_id(wf_id):
...
            workdir: pathlib.Path = typer.Argument(...,
            help='working directory for workflow containing input + output files',)):
     """Submit a new workflow."""
     if os.path.exists(wf_path):
-        wf_tarball = open(wf_path, 'rb')
+        # Check to see if the wf_path is a tarball or a directory. Run package() if directory
+        if os.path.isdir(wf_path):
+            print("Detected directory instead of packaged workflow. Packaging Directory...")
+            package(wf_path, pathlib.Path("/tmp"))
+            wf_tarball = open(pathlib.Path("/tmp/" + str(wf_path.name) + ".tgz"), 'rb')
```
pagrubel: Why place the tarball in /tmp. I'd either like to see it placed in the working dir specified by the submit or cwd. Another option would be to put it in bee_workdir/workflows/<wf_id>/.  It ends up there anyway. Right now it is left in /tmp

pagrubel (changes requested):
I'm thinking you should change the location of the tarball so that we don't have to do clean up.

beeflow/client/bee_client.py
pagrubel: @aquan9 After thinking about this using /tmp is fine, if you remove the tar ball once the workflow is submitted

jtronge:
You might be able to use [tempfile.TemporaryDirectory](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory).

beeflow/client/bee_client.py
aquan9: I was able to get it to be stored in bee_workdir/archives where I found other tarballs. The wf_id doesn't appear to get generated until after the submission which is too late in the code. I also tried the tempfile idea, but there are too many places where the code might fail where the tmp dir might need to be cleaned up quickly before submission, so the cleanest thing to do seemed to put it in bee_workdir/archives or some other bee_workdir/<directory> if you prefer something else.

aquan9:
I did a "git rebase develop" to pull in the documentation (and everything else that has happened in develop)

beeflow/client/bee_client.py
pagrubel: Thanks looks good. We may want to change the example a bit in the future. I'm going to add CWL to the command section it is such a small change and should have been stated differently to begin.
This all look great! Thanks, @aquan9

pagrubel (approved):

pagrubel (changes requested):
It seems the integration tests are failing

pagrubel (approved):
Looks good!

PR #589, opened 2022-11-09
Remove MARS and use Python 3.10 in CI

jtronge:
This will hopefully fix our Python version problems with GitHub Actions.


pagrubel:
There are still some references to mars in beeflow/scheduler/algorithms.py and beeflow/common/config_driver.py 
Do you think those should be cleaned up?

pagrubel:
@jtronge There are still some references to mars in beeflow/scheduler/algorithms.py and beeflow/common/config_driver.py
Do you think those should be cleaned up?

pagrubel (approved):

PR #599, opened 2022-11-17
Add docs workflow

jtronge:
This adds a documentation workflow that should install bee, run sphinx and then publish the HTML to github sites.


pagrubel (approved):
Need to approve and merge to test building the documentation

Issue #600, opened 2022-11-18
Fix Stderr Output


rstyd:
Stderr output isn't currently working. Need to fix it by modifying the slurm worker.

pagrubel:
satisfied by #605

PR #605, opened 2022-11-18
Add stderr to output in slurm worker

rstyd:
This PR addresses #600.


pagrubel (commented):
The reason tee was used was so the stdout would be in the archives as well as the place specified by the CWL. Now it is only in the output file specified by the stdout in CWL.
I think we want to preserve this and find a way to still separate the stderr and stdout.

pagrubel (changes requested):
We should still record outputs in the archives.

pagrubel:
Does moving build_text from worker.py to slurm_worker.py mean we also need a copy in lsf_worker.py and any new workers?

pagrubel:
I did some testing on this and now the mapping is if stderr is specified the application err output only goes to that path and the err output in the archive is from charliecloud. However if the application uses stderr output (such as ffmpeg) and only stdout is specified in CWL, the output from the application goes to that path and charliecloud output is the only thing in the archive. So my conclusion is that when using --output it combines stderr and stdout.

pagrubel (dismissed):

pagrubel:
Since for slurm the default is: when stdout is defined and stderr is not they are combined and this is true with this PR also.
So I'm going to approve the PR and we will document it for the user in another.

pagrubel:
Since this is behind develop and we want to use it as an example for a release I'm going to merge develop and test it first.

pagrubel:
I'm having trouble with checkpoint/restart example
```input:   tar       /yellow/users/pagrubel/.beeflow/container_archive/clamr-toss.tar.gz
output:  dir       /tmp/clamr-toss
analyzing ...
unpacking ...
done
srun: error: Could not open stdout file: Permission denied
srun: Job 438043 step creation temporarily disabled, retrying (Requested nodes are busy)
srun: Job 438043 step creation still disabled, retrying (Requested nodes are busy)
```
input:   tar       /yellow/users/pagrubel/.beeflow/container_archive/clamr-toss.tar.gz
output:  dir       /tmp/clamr-toss
analyzing ...
unpacking ...
done
srun: error: Could not open stdout file: Permission denied
srun: Job 438043 step creation temporarily disabled, retrying (Requested nodes are busy)
srun: Job 438043 step creation still disabled, retrying (Requested nodes are busy)
srun: error: Unable to create step for job 438043: Job/step already completing or completed
srun: error: Unable to create step for job 438043: Job/step already completing or completed

pagrubel:
So I guess checkpointing is messed up. We can go forward with your PR and make an issue for it, but I'd like to see develop merged into the branch then I'll change pyproject.toml to version 0.1.2 so we can procede with publishing tests. Does that seem reasonable?

Also we may need to think about saving separate output files for checkpoint tasks.

pagrubel (approved):

PR #611, opened 2022-11-18
Fix --version for release

jtronge:
This should fix `beeflow --version` for the release.


pagrubel (approved):
Needed to increase to 0.1.1 for pip installation (cannot re use versions)

Issue #318, opened 2021-04-13
Look over dependencies 


rstyd:
We currently have an extremely large number of dependencies. I will look over what we currently have and see if there is anything we can cut.

pagrubel:
If this doesn't prevent use of BEE it can be done after the release @rstyd I'm add the label.

PR #619, opened 2022-11-29
Update and remove extra dependencies

jtronge:
This should address issue #318. I've updated Flask and Jinja and removed a couple dependencies that we don't seem to be using right now.


rstyd (approved):
Looks great!

pagrubel (approved):
Tried this out by clearing out poetry environment and reinstalling and running an example on darwin

PR #625, opened 2022-12-14
Add enhanced/gui client

jtronge:
There is a lot here so I thought it was a good idea to open this now. This is not completely working yet so I'm marking this as WIP, but I don't think I'll make a lot of major changes to it.


jtronge:
I'm going to close this since I moved back to an earlier version of the code and created a new branch.

PR #642, opened 2023-01-06
GUI Client

jtronge:
There is still a lot that needs to be done with this, but I'm opening it as a WIP so that people can see what state it's in.

This can submit workflows, but it currently doesn't display any workflow information or allow for start/stop. There are also many other issues, including listing multiple workflows, managing multiple connections, etc. I've also had difficulties getting errors back when an SSH tunnel fails (for example one might have misspelled a username or something else).


jtronge:
I ended up reducing most of the code down to a very simple interface which just allows you to view a running workflow. This requires you to enter the system hostname, your username and the bolt port for the given workflow. It then opens up an SSH tunnel for just the bolt port and displays the workflow.

Screen Shot 2023-01-11 at 11 20 18 AM

To get the bolt port for a workflow, you can now run `beeclient metadata ${wf_id}` which will display it (and could also be used for other information later on).

Note that this doesn't include any support for workflow submission or viewing workflows. I think this can be done, but we'll need more time to update the code for the current BEE and ensure there aren't any major problems to be fixed.

pagrubel:
@jtronge Did you mean to include binaries (tar files)

jtronge:
No I didn't. Those should be removed now.

pagrubel (changes requested):
Should the tar balls in the enhanced_client folder be deleted? oops I see they are deleted now

pagrubel (approved):
Viz works well until the workflow is completed. Good job. I think we should merge this and have an issue for using the newer database to in order visualize completed workflows.

PR #645, opened 2023-01-10
Ml workflow

pagrubel:
Just testing


pagrubel:
I wanted to add Ragini's machine learning cwl files to our collection. They were part of cwl validation work and we shouldn't lose them. They should be in our history.

rstyd (approved):
LGTM

PR #646, opened 2023-01-11
Fix PyLama and poetry install errors

jtronge:
This should hopefully fix the pylama error in CI.

Sphinx is still causing `poetry update` to loop forever, so it's commented out for now, but we still need to figure out what version or dependency is causing that problem.


pagrubel:
Poetry update works for version 1.3.2

pagrubel (approved):
Looks good, thanks.
We may want to look at updating to newer versions as they come out.

PR #648, opened 2023-01-27
Fixes for Newer Slurm Versions

jtronge:
Previously the `slurm_worker.py` code had the slurm openapi version hard coded. This moves the version info out into an option in the config so that we can support newer versions of Slurm.


pagrubel (approved):
Looks good and works on crusher, thanks!

Issue #643, opened 2023-01-06
Single file CWL broken input


jtronge:
I think that single-file CWL workflows are currently broken. I noticed this when trying out the `lulesh-mpi` workflow, which gives an error: `Submit: Parser: required input lulesh/iterations not satisfied`, when in fact this input exists. When I moved the CommandLineTool listing out into another file it submitted without error.

See <https://github.com/lanl/BEE/tree/e7ffab6d4b4dbfeef9b472a85aaa67fa699b5016/beeflow/data/cwl/bee_workflows/lulesh-mpi> for the workflow.

PR #651, opened 2023-02-03
Modify code in the worker and job templates to ensure that MPI works

jtronge:
MPI was working for one-node jobs, but not for multi-node jobs. This updates the worker and job template code to make sure that `ch-convert` gets called correctly for all job types.

I tested the `comd-mpi` and `lulesh-mpi-multi-file` workflows under `beeflow/data/cwl/bee_workflows/`. Note that the `lulesh-mpi` workflow seems to be broken for an unrelated reason (see issue #643).

I think it might be a good idea to refactor the worker and container runtime code. We could also make the jinja template only internal, so the user would never have to touch it. Although then we would need to gradually add features as users needed them.


pagrubel:
If a new user runs beecfg new and selects the default jinja they will get a batch job that only runs for 10 seconds. I can only imagine you put this in there so that we can design some ci or other tests for checkpointing. If we are going to put that in the default jinja template, I'd suggest picking up the system default. However for now i don't think it should be in there.

pagrubel (approved):
MPI worked well on crusher and darwin. On Crusher (and most likely fog) one needs to use absolute names for the working directory when submitting workflows.

Also non-mpi workflows still work! Thanks for the update @jtronge

Issue #622, opened 2022-12-01
Merge submit and start client commands into one


jtronge:
The new merged command should return the workflow ID immediately, spawning some sort of background thread to launch neo4j and then start the workflow.

jtronge:
One problem with this is that the user won't see if the parser failed immediately. They would have to check the status or look at the logs to catch an error.

pagrubel:
Could we have a no-start option and keep a start option?

pagrubel:
Something like:
submit (automatically starts)
submit --no-start

pagrubel:
Feature added in #653

PR #653, opened 2023-02-13
Merge submit and start commands

jtronge:
This should close issue #622. By default `beeclient submit ...` should also start the workflow, unless `--no-start` has been passed.


pagrubel:
@jtronge Thanks this looks great. Should we correct user documentation in this PR or submit another since it does change functionality.

pagrubel:
@jtronge This looks great!

Is there a way to add an explanation for the option ``--no-start``  to the message you get for  ``beeclient --help ``
Also, the conflict needs to be fixed.

jtronge:
Ok, the merge conflicts should be fixed.

I think it should show an explanation when you run `beeclient submit --help`, under the Options heading.

pagrubel (approved):

Issue #656, opened 2023-02-16
beecfg doesn't use updated slurm-submit jinja file


rstyd:
`beecfg new` currently isn't using the new slurm-submit.jinja file. This causes a `submit-fail` when submitting jobs.

PR #657, opened 2023-02-16
Fix submit jinja filename

rstyd:
Addresses #656.


jtronge (approved):
Thanks, looks good.

pagrubel (approved):
Nice this also fixed some intermittent problem I was having on Crusher (intermittent because I had some script that would rewrite those configuration files)

Issue #635, opened 2022-12-21
Look into alternative to slurmrestd so that it won't be a requirment on systems


pagrubel:
When requesting slurmrestd on Crusher we ran into problems with the version of openapi
The admins made some suggestions that we should look into using such that slurmrestd is not required.

The colleague who shared the example also mentioned that `squeue --json` might actually be more convenient, and it uses the same backend code to generate the json as the `slurmrestd` daemon. It's what he used to write the `/usr/bin/jobstat` tool. It sounds like your system already uses the daemon, but just in case you want to look into it...

 

Please let me know if this will allow you to move forward for now, or if there are other capabilities you need that this doesn't address. Thanks.

 

-Tom

 

[crusher: ~]$ jobstat -u tpapathe
...
JobID      User         Account            Nodes Walltime   SubmitTime     Priority JobName
228886     tpapathe     stf016             1     20:00      12/13_17:26:53 200      interactive
 
 
[crusher: ~]$ echo -e "GET /slurm/v0.0.37/job/228886 HTTP/1.1\r\n" | /usr/sbin/slurmrestd
slurmrestd: operations_router: [fd:0->/dev/pts/8] GET /slurm/v0.0.37/job/228886
slurmrestd: rest_auth/local: slurm_rest_auth_p_authenticate: [fd:0->/dev/pts/8] accepted connection from user: tpapathe[5987]
HTTP/1.1 200 OK
Content-Length: 3586
Content-Type: application/json{
   "meta": {
     "plugin": {
       "type": "openapi\/v0.0.37",
       "name": "Slurm OpenAPI v0.0.37"
     },
     "Slurm": {
       "version": {
         "major": 22,
         "micro": 5,
         "minor": 5
       },
       "release": "22.05.5"
     }
   },
   "errors": [
   ],
   "jobs": [
     {
       "account": "stf016",
       "accrue_time": 1670970413,
       "admin_comment": "",
       "array_job_id": 0,
       "array_task_id": null,
       "array_max_tasks": 0,
       "array_task_string": "",
       "association_id": 114,
       "batch_features": "",
       "batch_flag": false,
       "batch_host": "",
       "flags": [
       ],
       "burst_buffer": "",
       "burst_buffer_state": "",
       "cluster": "crusher",
       "cluster_features": "",
       "command": "",
       "comment": "",
       "contiguous": false,
       "core_spec": null,
       "thread_spec": null,
       "cores_per_socket": null,
       "billable_tres": null,
[...]

PR #660, opened 2023-02-28
Slurm command line worker

jtronge:
I think this should address issue #635. It renames the slurmrestd config section to `slurm` and adds a new `use_commands` option that gets set by beecfg.

I'd like to test this somehow, although it probably wouldn't be feasible to do in CI without adding too much time.

### DIFF
--- a/beeflow/cli.py
+++ b/beeflow/cli.py
@@ -24,6 +24,9 @@ from beeflow.common.config_driver import BeeConfig as bc
 from beeflow.common import cli_connection
 
 
+bc.init()
+
+
 class ComponentManager:
     """Component manager class."""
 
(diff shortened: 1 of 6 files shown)


pagrubel:
Is there a way to set up a test and run a github action before merge, even if it is manual?
Then we could change our procedures to make sure that is done before merging.

jtronge:
Yeah I could edit the CI scripts one time and remove the WIP.

pagrubel:
That may be okay for getting the PR merged but not practical for ensuring future merges don't fail

pagrubel:
@jtronge Let's figure a way to test it manually on darwin and then we can set up some nightly tests

pagrubel:
I tested this on darwin by removing the path to slurmrestd from $PATH and rerunning beecfg new, and it works fine. I still don't know how we would make this an automated test though.
@jtronge What do you think about making the "Not using slurmrestd" red or some color?

pagrubel:
@jtronge for CI would it be too expensive to do the following to test using commands instead of slurmrestd: 
After the current integration test stop beeflow; change the value of use_commands to True in bee.conf and re run the tests?  It won't test the configuration but atleast if the commands change for some reason BEE won't break

jtronge:
@pagrubel I think I found a way to make the slurm worker unit tests run with commands as well.

If you run it with `pytest beeflow/tests/test_slurm_worker.py -v`, it should show tests run with/without slurmrestd.

pagrubel (approved):

Issue #628, opened 2022-12-15
Add a way to specify the dependency container from the command line


pagrubel:
Since we have to answer the question the user cannot use auto complete to find the dependency container. If we add a way to specify it on the command line with beecfg such as:
``beecfg new Slurm <path to dependency container> ``
One would not need to answer the questions.

pagrubel:
Address by #664

PR #664, opened 2023-03-15
Improve usability of 'beecfg new'

jtronge:
I think this should fix issue #628. Rather than taking those options on the command line you should be able to tab complete file names now. This only works for the `bee_dep_image` option, but I think that's all we need (and the tab completion code is there if we need it for something else).


pagrubel (approved):
Thanks!

PR #670, opened 2023-03-29
Add restart code to beeflow

jtronge:
Beeflow should now restart components on failure. I've added a `max_restarts` options as well.


rstyd (approved):
Works great! Approved.

PR #671, opened 2023-04-03
Fix charliecloud --home issue

rstyd:
This PR hotfixes an issue where the charliecloud CRT is unable to access the home directory in images since charliecloud no longer bindmounts home by default.


rstyd:
Need to discuss this issue more.

Issue #631, opened 2022-12-21
Checkpoint/Restart is broken


pagrubel:
When testing the latest PR it seems that the checkpoint/restart was not working. 
We need to explore what is happening  and also create a test for it even if it is just to run manually.
Also we need to consider what to do with outputs stdout and stderr that may already exist for steps that are created due to checkpointing.

jtronge:
For stdout/stderr maybe we should append to the output file. We might also be able to do something based on a numbering scheme (like clamr-stdout-0.txt, clamr-stdout-1.txt, ...) but that might be a little more complicated to handle.

Right now I think each restart task will overwrite the previous output. I think if we add `--open-mode=append` as an sbatch directive it might work.

jtronge:
This should have been fixed in PR #680.

Issue #678, opened 2023-04-14
Builder broken with latest Charliecloud


jtronge:
I'm getting this error when trying to build with the latest charliecloud version: `ch-image build: error: argument --force: invalid choice: '/tmp' (choose from 'fakeroot', 'seccomp')`.

jtronge:
I think PR #680 should have fixed this. We updated the min Charliecloud version to 0.32.

PR #680, opened 2023-04-18
Checkpoint restart fix and test script

jtronge:
This should fix issues #631 and #678.

The recent changes in Charliecloud don't seem backward compatible, so I've updated the min version to 0.32. I'm wondering if that's the best way to deal with this?


jtronge:
A couple things that I forgot to mention:

This adds a new requirement for setting a time limit:

```
beeflow:SchedulerRequirement:
  timeLimit: 00:00:10
```

It also adds a new beeflow DockerRequirement extension, `beeflow:forceType`, which corresponds to the newly required argument for ch-image's `--force` option.

Also I've fully "internalized" the jinja file, so that it no longer gets copied to the user's config dir.

pagrubel:
Just starting to look at this. How would one enter other directives such as account as required on Summit, really any system where one has multiple accounts?

jtronge:
We would probably want to add that as an option under the `beeflow:SchedulerRequirement`. Maybe we could also set a default in the bee.conf in case somebody will always be using the same one.

pagrubel:
> We would probably want to add that as an option under the `beeflow:SchedulerRequirement`. Maybe we could also set a default in the bee.conf in case somebody will always be using the same one.

It would need to be done in this PR or BEE becomes unusable on the ORNL platforms

jtronge:
I added an `account` option for the `beeflow:SchedulerRequirement` as well as a new bee.conf section `[job]` for setting default account and timeout options. I also tested this on Crusher.

Also, to run the checkpoint restart test, first start beeflow and then run `python3 ci/integration_test.py --tests checkpoint_restart --timeout 800`.

rstyd (approved):
Looks good to me!

PR #687, opened 2023-05-09
Add partition option for beeflow:SchedulerRequirement

jtronge:
This should allow for someone to specify a partition in a workflow.


pagrubel:
Nice I tested using the default in the bee.conf and adding a hint for a different partition in one step.

pagrubel (approved):

PR #696, opened 2023-06-26
Flux support

jtronge:
This should add initial flux support to bee along with a pennant scalability workflow.


jtronge:
I've added some initial CI testing for flux. It's taking about 14 minutes to run (versus about 8 for the Slurm version).

pagrubel:
This looks good. The example needs to be scrubbed some. You shouldn't refer to actual cluster it was tried on and if you are going to keep the example in the examples directory, some instructions should be included in the documentation.

examples/pennant-rzvernal/pennant.yml
```diff
@@ -0,0 +1 @@
+pnt: 'sedovbig.pnt'
```
pagrubel: change this to '/PENNANT/test/sedovbig/sedovbig.pnt'
So the user won't have to copy the file to their workdir

examples/pennant/sedovbig.pnt
pagrubel, line 1: This file isn't necessary just access the one in the container through the yml file.

pagrubel (changes requested):
If you leave the example in examples, write up some documentation. At this point I'm leaning towards putting it in our own cwl directory, since it isn't a true scaling study, or state that it isn't a performance test in the documentation.

You might change out your username in the path etc and get rid of references to rzvernal.

jtronge:
I scrubbed the pennant workflow and moved it to the beeflow data directory.

There seems to be a strange dependency error happening in CI and I don't think it's related to my changes. I'm going to try a couple updates to fix that.

jtronge:
I updated the PyYAML version and this seems to have fixed the CI tests. Looks like they just made a new release on July 17, although I'm not sure if this is exactly what caused poetry problems.

pagrubel (approved):

PR #697, opened 2023-07-06
Move parser to client

### LINKED ISSUE #667: Move parser to client
This will enable us to validate the workflow at the beginning and give knowledge about the files the workflow will use so that we don't archive extra files.

Boogie3D:
The parser is now invoked by the client _before_ sending data to the WF Manager. Resolves #667 

The following changes were made:

- `CwlParser` is now instantiated in `bee_client.py`
- `CwlParser` no longer invokes Workflow Interface, now returns `Workflow` and `Task` objects
- `WorkflowInterface.initialize_workflow()` now takes only `Workflow` and `WorkflowInterface.add_task` now takes only `Task`
- `jsonpickle` is used to encode the `Workflow` and `Task` objects and send them to the WF Manager via REST
  - The objects are decoded and reconstructed by the WF Manager and then uploaded to the graph database
- CWL filenames are no longer sent to the WF Manager via REST
- `test_parser.py` and `test_wf_interface.py` have been updated and are passing


rstyd:
Approved!

pagrubel:
@rstyd @Boogie3D The merge of the flux PR causes conflicts here so please fix

pagrubel (approved):

Issue #701, opened 2023-07-25
Make it easier to change the bee workdir in the config


jtronge:
Right now the `~/.beeflow` path is duplicated in the config file for a number of options. This should be set one time in some sort of variable and then referenced with something like `${bee_workdir}/...`, or those specific options should be removed and just default to using the bee workdir as set in the config.

pagrubel:
Completed in #709

PR #709, opened 2023-07-28
Remove bee_workdir duplication in config file

jtronge:
I think this should fix issue #701. I've removed most the redundant options in the config file to make it easier for someone to change the `.beeflow` path.


pagrubel (approved):
Looks good to me.

Issue #706, opened 2023-07-26
Combine Beeflow and Beeclient


rstyd:
It's kind of confusing having two different commands so we should combine them into one. 

One issue we'll run into is that there are several commands that are used in both such as start and status. Our current working solution is to put the beeflow commands behind `beeflow core {start|stop|status}` since the client commands will be used more often. 

In order to keep the code clean, we'll want to keep the beeclient code exactly where it is and just combine the typer interface for both applications.

PR #710, opened 2023-08-08
Combine beeflow, beeclient, and beecfg

jtronge:
Addresses issue #706, moving `beeflow` to `beeflow core` and `beecfg` to `beeflow config`.

There's also the `beecloud` command, but this uses argparse instead of typer for argument handling and I suspect that much of that code has fallen behind the other changes we've made in the past year or so.


pagrubel:
This looks good to me. I'll wait for @rstyd to review

rstyd (dismissed):
Approved!

rstyd (approved):
LGTM!

PR #713, opened 2023-08-22
Fix Dependencies Bug

Boogie3D:
This PR fixes a bug where dependencies were not properly constructed in Neo4j when a task was added after any tasks that depend on it. This was caused by a partially ineffective Cypher query, which has now been split into two correctly functioning queries.


pagrubel (approved):
Thanks, this works and looks good!

Issue #720, opened 2023-09-06
Using a pre-release version of Charliecloud causes an error


pagrubel:
If someone is attempting to use the most current development version of Charliecloud from github they will encounter an error with beeflow:
```
(hpc-beeflow-py3.9) grubel2@rzvernal10 BEE$ ch-run --version
0.34~pre+5d8f4c2
(hpc-beeflow-py3.9) grubel2@rzvernal10 BEE$ beeflow start
Checking dependencies...
Traceback (most recent call last):

  File "/g/g11/grubel2/.cache/pypoetry/virtualenvs/hpc-beeflow-jTxkSGXx-py3.9/bin/beeflow", line 6, in <module>
    sys.exit(main())

  File "/g/g11/grubel2/BEE/beeflow/cli.py", line 420, in main
    app()

  File "/g/g11/grubel2/.cache/pypoetry/virtualenvs/hpc-beeflow-jTxkSGXx-py3.9/lib/python3.9/site-packages/typer/main.py", line 289, in __call__

  File "/g/g11/grubel2/.cache/pypoetry/virtualenvs/hpc-beeflow-jTxkSGXx-py3.9/lib/python3.9/site-packages/typer/main.py", line 280, in __call__

  File "/g/g11/grubel2/.cache/pypoetry/virtualenvs/hpc-beeflow-jTxkSGXx-py3.9/lib/python3.9/site-packages/click/core.py", line 1130, in __call__

  File "/g/g11/grubel2/.cache/pypoetry/virtualenvs/hpc-beeflow-jTxkSGXx-py3.9/lib/python3.9/site-packages/click/core.py", line 1055, in main

  File "/g/g11/grubel2/.cache/pypoetry/virtualenvs/hpc-beeflow-jTxkSGXx-py3.9/lib/python3.9/site-packages/click/core.py", line 1657, in invoke

  File "/g/g11/grubel2/.cache/pypoetry/virtualenvs/hpc-beeflow-jTxkSGXx-py3.9/lib/python3.9/site-packages/click/core.py", line 1404, in invoke

  File "/g/g11/grubel2/.cache/pypoetry/virtualenvs/hpc-beeflow-jTxkSGXx-py3.9/lib/python3.9/site-packages/click/core.py", line 760, in invoke

  File "/g/g11/grubel2/.cache/pypoetry/virtualenvs/hpc-beeflow-jTxkSGXx-py3.9/lib/python3.9/site-packages/typer/main.py", line 607, in wrapper

[...]

PR #722, opened 2023-09-06
Fix charliecloud version code for pre-release versions

rstyd:
This PR addresses #720.


pagrubel:
@rstyd I committed a change to actually print the pre-release version if that is what is being used. Please look it over and let me know if that's fine with you.

pagrubel (approved):

Issue #705, opened 2023-07-25
Remove delay time after workflow submit


rstyd:
When we submit a workflow, there is a delay while the workflow manager waits for the neo4j instance to finish loading. We should offload the wait to a subprocess and instantly return a workflow ID to the user after we parse the workflow in the client.

pagrubel:
I still think there is the possibility of someone wanting to submit another workflow immediately. Since you will most likely create a new workflow state (Submitting) so I guess we could check for that state before attempting another and a message could come back, "Another workflow is in Submitting state, please submit when it finishes submission" or something like that.)

rstyd:
If someone submits another workflow immediately, it won't be a problem since the previous workflow submission will be worked on by another process and we'll just fork a new processes for the new workflow.

pagrubel:
Fixed in #728

Issue #702, opened 2023-07-25
Command to Download Default Dependencies Container


Boogie3D:
Rather than simply attempting to point users toward a usable BEE dependencies container, there could be a client command to download the container from DockerHub, etc using Charliecloud's `ch-image pull` command. We should decide whether we want to pull this into Charliecloud's local image repo and keep it there or convert it into a tarball and store on the filesystem. This method may not be usable on restricted networks, but those circumstances can be handled on a case-by-case basis by providing the appropriate environment on those machines.

Possible commands are:
- `beeclient pull`
- `beeclient pull-deps`
- `beeclient download-deps [dest_dir]` (default: `.`)
- `beeclient dependencies`

pagrubel:
Could we have a github action that builds the container and push it to dockerhub. We would run this manually.

pagrubel:
Addressed in PR #728

PR #728, opened 2023-10-02
Remove delay time on workflow submit

jtronge:
This should address issue #705. Tasks to start the GDB and initialize workflows are now offloaded to a celery backend based on Redis. I'm marking this as WIP since we still need to figure out the containers to use.

### DIFF
--- a/beeflow/client/core.py
+++ b/beeflow/client/core.py
@@ -193,6 +193,7 @@ def init_components():
     def redis():
         """Start redis."""
         data_dir = 'data'
+        os.makedirs(os.path.join(paths.redis_root(), data_dir), exist_ok=True)
         conf_name = 'redis.conf'
         container_path = paths.redis_container()
         # If it exists, we assume that it actually has a valid container
--- a/beeflow/client/core.py
+++ b/beeflow/client/core.py
@@ -194,6 +194,8 @@ def init_components():
         """Start redis."""
         # Dump the config
         with open(paths.redis_config(), 'w', encoding='utf-8') as fp:
+            # Don't listen on TCP
+            print('port 0', file=fp)
             print('dir', paths.redis_data(), file=fp)
             print('maxmemory 2mb', file=fp)
             print('unixsocket', paths.redis_socket(), file=fp)
--- a/beeflow/client/core.py
+++ b/beeflow/client/core.py
@@ -449,3 +449,26 @@ def restart(foreground: bool = typer.Option(False, '--foreground', '-F',
     """Attempt to stop and restart the beeflow daemon."""
     stop()
     start(foreground)
+
+
+def pull_to_tar(ref, tarball):
+    """Pull a container from a registry and convert to tarball."""
+    subprocess.check_call(['ch-image', 'pull', ref])
+    subprocess.check_call(['ch-convert', '-i', 'ch-image', '-o', 'tar', ref, tarball])
+
+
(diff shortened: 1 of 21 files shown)


jtronge:
This adds a new config option `redis_image` for the Redis container. Right now it's just using the official one from docker hub.

jtronge:
The documentation for the new redis container should be updated now. I also changed `bee_dep_image` to `neo4j_image` in the config to be consistent.

jtronge:
I just added a command `beeflow core pull-deps` to download both the Neo4j and Redis containers. I think this should address #702.

pagrubel (changes requested):
Should the command really be ``beeflow config pull-deps``?
Also, can you use the code that is used to load the charliecloud module if it doesn't exist to load it or warn that charliecloud is not available when using the command with the url for charliecloud, otherwise the user gets an error saying ch-image does not exist.
And the command to download the containers should also be documented in the command line interface section of either core or config wherever it ends up.

This probably should have been a separate pull request. I still need to review the rest of the pull request.

pagrubel:
I'm getting an error when I try to run the simple cat-grep-tar example on darwin. Attaching the error log. 
[bee-error-1697478790.log](https://github.com/lanl/BEE/files/12920306/bee-error-1697478790.log)

pagrubel:
> I'm getting an error when I try to run the simple cat-grep-tar example on darwin. Attaching the error log. [bee-error-1697478790.log](https://github.com/lanl/BEE/files/12920306/bee-error-1697478790.log)

Deleted my bee_workdir to have a clean copy and this cleared up the problem.

beeflow/client/core.py
```diff
@@ -181,6 +181,47 @@ def start_scheduler():
...
+        data_dir = 'data'
+        os.makedirs(os.path.join(paths.redis_root(), data_dir), exist_ok=True)
+        conf_name = 'redis.conf'
+        container_path = paths.redis_container()
+        # If it exists, we assume that it actually has a valid container
+        if not os.path.exists(container_path):
+            subprocess.check_call(['ch-convert', '-i', 'tar', '-o', 'dir',
+                                   bc.get('DEFAULT', 'redis_image'), container_path])
+        # Dump the config
+        with open(os.path.join(paths.redis_root(), conf_name), 'w', encoding='utf-8') as fp:
```
rstyd: I think we should only dump a config if one doesn't currently exist.

beeflow/wf_manager/resources/wf_actions.py
```diff
@@ -22,20 +22,12 @@ def __init__(self):
...
-                                         status='ok'), 200)
-            return resp
-        wfi.execute_workflow()
-        tasks = wfi.get_ready_tasks()
-        wf_utils.schedule_submit_tasks(wf_id, tasks)
-        wf_id = wfi.workflow_id
-        wf_utils.update_wf_status(wf_id, 'Running')
-        db.workflows.update_workflow_state(wf_id, 'Running')
-        resp = make_response(jsonify(msg='Started workflow!', status='ok'), 200)
+        if wf_utils.start_workflow(wf_id):
```
rstyd, line 25: This is a great change!

rstyd (approved):
Everything looks great to me. I approve!

pagrubel (approved):
Thanks Jake this is great!

PR #732, opened 2023-10-24
Make task manager background time interval configurable

jtronge:
This makes the time interval in the task manager configurable, currently with a default of 10 seconds. This should address issue #741.


pagrubel (approved):
Looks good, thanks!

PR #738, opened 2023-10-31
Fix issues with leaking gdb processes

### LINKED ISSUE #730: Cancel workflows leaves an orphaned neo4j instance
The neo4j pid should be in the wf manager database. The neo4j instance should be killed if a workflow is cancelled.

Also, should the workflow be left in the database with a CANCELLED status?

aquan9:
Currently we are aware of two pathways to a leaked gdb process: A failed workflow, and a cancelled workflow. 

Previously, failed workflows were detected in the workflow manager, but left unhandled.
Previously, cancelled workflows were simply deleted from the database. 

Now, cancelled workflows are marked as cancelled, in the db, and so evidence of the workflow's existence remains when you query `beeflow list`. In processing the workflow cancellation we also cleanly stop the gdb now. 

The case of the failed workflow was also handled, and is now cleanly archiving and stopping the gdb upon failure. 

One thing to note for a future issue, is that both completed workflows and failed workflows results in a final "archived" state. This can be misleading for the user. I'll go ahead and make an issue for this. 

Should fix #730 
I also went ahead and added an example workflow that guarantees a workflow failure (via an invalid command), for the purpose of testing.


pagrubel:
Please fix the failing unit test.

Issue #737, opened 2023-10-31
Celery worker starting too many processes


jtronge:
On running `beeflow core start`, celery seems to be starting upwards of 30 processes that could potentially cause problems on multi-user systems. This may just be a problem with the configuration that we need to look into.

pagrubel:
Fixed in #739

Issue #736, opened 2023-10-31
Beeflow core start


pagrubel:
Should we  wait and give a status

1. Why does it take so long to start everything.
2. If that can be shortened, wait and give status.

pagrubel:
Closing this since the time was shortened in  #739 so status can be checked immediately using ``beeflow core status``

PR #739, opened 2023-11-01
Improve start speed and decrease celery processes

jtronge:
I think this should address #736 and #737. The Redis container is now unpacked before the core goes into the background, and celery should only be launching a single process.


pagrubel (approved):
This is much better, no extra celery process and runs faster so when I start beeflow core, I can immediately turn around and issue ``beeflow core status``

Issue #711, opened 2023-08-18
Can we use files that are not in the directory for the workflow such as a different yml file?


pagrubel:
We need to clarify how this works. In the documentation it says:
```Additionally, if the main_cwl and yaml files are not in the workflow directory, they will be copied into a temporary copy of the workflow directory before packaging. Compare this with the previous example. ```

When I tried to use a main_cwl file and a yml file that isn't in the directory specified on the submit line, it couldn't find the step cwl files.
I am going to change the documentation so the example works, but we should decide what is happening here and show a use case.

**Comment by pagrubel** [comment]
So, I tested this out, apparently we can use a yml input file that is not in the workflow directory, however the main file we use needs to be the one in the main workflow dir.
I get this error when trying to use a different main cwl which maybe doesn't make sense anyway, however we should handle the error better. Notice I'm trying to use the clamr-ffmpeg-build directory which has all the files for the workflow, but select the clamr_wf.cwl in the current directory. 

```
ll
total 23
drwxrwxr-x 2 pagrubel pagrubel  4096 Aug 18 17:02 clamr-ffmpeg-build
-rw-rw-r-- 1 pagrubel pagrubel   404 Aug 21 13:56 clamr_job.yml
-rw-rw-r-- 1 pagrubel pagrubel  1882 Aug 21 13:55 clamr_wf.cwl
-rw-r--r-- 1 pagrubel pagrubel  4791 Aug 21 15:03 ffmpeg_stderr.txt
drwxrwxr-x 2 pagrubel pagrubel  4096 Aug 21 15:03 graphics_output
-rw-rw-r-- 1 pagrubel pagrubel  3215 Aug 18 16:49 lorem.txt
-rw-r--r-- 1 pagrubel pagrubel   302 Aug 21 13:52 occur0.txt
-rw-r--r-- 1 pagrubel pagrubel   229 Aug 21 13:52 occur1.txt
-rw-rw-r-- 1 pagrubel pagrubel 10240 Aug 21 13:53 out.tgz
-rw-rw-r-- 1 pagrubel pagrubel    64 Aug 21 15:03 total_execution_time.log
(hpc-beeflow-YDRVf3zF-py3.9) (base) pagrubel@darwin-fe1 beeworkdir2$ beeflow submit clamrb clamr-ffmpeg-build clamr_wf.cwl clamr_job.yml ~/beeworkdir2
Detected directory instead of packaged workflow. Packaging Directory...
Traceback (most recent call last):

[...]

PR #743, opened 2023-11-14
Issue#711/fix alternate main cwl

pagrubel:
This addresses issue #711 by parsing the workflow after moving the wf cwl directory to a tmp directory and copying the alternate files in then packaging and parsing the new package. Previously, the workflow was parsed directly from the workflow path before copying the alternate files.


jtronge (approved):
I tested this with the cat-grep-tar workflow, moving the input and main cwl to different places, and it seems to work.

PR #749, opened 2023-12-01
Stop slurm tests from creating tempfiles in $HOME

jtronge:
I noticed that the slurm worker tests were creating random temp files in $HOME. This should hopefully stop that from happening.


pagrubel (approved):
Thank you!

Issue #704, opened 2023-07-25
Restart GDB on Failure


Boogie3D:
We should be able to detect when either the GDB process has stopped running or when the GDB container itself has stopped running and restart the container in either case. Main questions are:

- How to detect?
  - WorkflowInterface throws an error when it can't connect to the Neo4j interface. Is this sufficient or should we detect some other way?
- Which component(s) should be responsible for detection and for restarting?
- If somehow the container is still running but not the Neo4j process, should we just restart the process without restarting the container? Is this situation possible?

pagrubel:
Addressed in #759

PR #759, opened 2023-12-18
Restart GDB

rstyd:
This PR addresses #704.

### DIFF
--- a/beeflow/common/db/wfm_db.py
+++ b/beeflow/common/db/wfm_db.py
@@ -110,10 +110,6 @@ class Workflows:
         stmt = "UPDATE tasks SET state=? WHERE task_id=? AND workflow_id=?"
         bdb.run(self.db_file, stmt, [state, task_id, workflow_id])
 
-    def update_gdb_pid(self, gdb_pid, workflow_id):
-        stmt = "UPDATE workflows SET gdb_pid=? WHERE workflow_id=?"
-        bdb.run(self.db_file, stmt, [gdb_pid, workflow_id])
-
     def get_tasks(self, workflow_id):
         """Get all tasks associated with a particular workflow."""
         stmt = "SELECT * FROM tasks WHERE workflow_id=?"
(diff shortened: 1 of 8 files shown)


beeflow/common/db/wfm_db.py
```diff
@@ -53,7 +53,7 @@ def __init__(self, db_file):
         """Initialize Task, db_file, and Workflow object."""
         self.Task = namedtuple("Task", "id task_id workflow_id name resource state slurm_id") #noqa
         self.db_file = db_file
-        self.Workflow = namedtuple("Workflow", "id workflow_id name state run_dir bolt_port gdb_pid") #noqa
+        self.Workflow = namedtuple("Workflow", "id workflow_id name state run_dir bolt_port http_port, https_port, gdb_pid") #noqa
```
jtronge: Should commas be between all of those? I guess it's just stripping out the extra commas for http_port and https_port.

beeflow/wf_manager/resources/wf_utils.py
```diff
@@ -123,16 +125,42 @@ def create_wf_namefile(wf_name, wf_id):
...
+            # Returns an exception if the GDB is running
+            os.kill(gdb_pid, 0)
+            # The GDB process is running so we'll kill it and restart it
+        except ProcessLookupError:
+            # The GDB process isn't running. Restart it
+            log.error('The GDB is currently down restarting it')
+            wf_dir = get_workflow_dir(wf_id)
+            http_port = db.workflows.get_http_port(wf_id)
+            https_port = db.workflows.get_https_port(wf_id)
+            gdb_pid = dep_manager.start_gdb(wf_dir, bolt_port, http_port, https_port)
```
jtronge: Will this get activated by a user request (i.e. a beeflow query)? I wonder if we should launch this as a celery task.

jtronge:
I tried this with `examples/clamr-ffmpeg-build`. I killed neo4j while the clamr task was running and it looks like the WFM restarted neo4j. I think the clamr task completed but the state is still at RUNNING and it doesn't get to the ffmpeg step.

rstyd:
Oh I'll check that out. Thanks! I bet the code that normally sets that is getting short skipped.

pagrubel:
> I tried this with `examples/clamr-ffmpeg-build`. I killed neo4j while the clamr task was running and it looks like the WFM restarted neo4j. I think the clamr task completed but the state is still at RUNNING and it doesn't get to the ffmpeg step.



> Oh I'll check that out. Thanks! I bet the code that normally sets that is getting short skipped.

You may also need to look at what commands you are using to check the status. I noticed in the past if a job finishes the status might not be available on some systems.

pagrubel:
@rstyd There have been several merges into develop so please merge develop in this branch

jtronge:
This works for me. I tried the clamr-ffmpeg-build workflow and killed the gdb twice, during the clamr and then the ffmpeg step, and both times it was restarted.

@rstyd Can you merge/rebase on top of develop to check against the recent neo4j updates? I'll approve after that.

pagrubel:
@rstyd I also tried this and it works for me. Would you please make a squash commit with the important description of what was done. There have been many changes to develop since you started.

rstyd:
No problem!

pagrubel:
Also, reexcute works well, thanks for fixing that!!!

jtronge (approved):
Looks great! I tested the clamr workflow and bee had no problems restarting neo4j.

PR #760, opened 2024-01-03
Add new checkpoint-restart tests and refactor integration test code

jtronge:
This is a WIP but I'm not adding the label so that the tests will run in CI.

I've added a new checkpoint-restart test that checks for a failure after `num_tries` is exceeded. I also decided to try moving the integration code from the script in `ci/` into `beeflow` to make it easier to work with.


jtronge:
The checkpoint-restart tests are not run by default, but can be started with `./ci/integration_test.py -t checkpoint_restart,checkpoint_restart_failure` after starting beeflow.

pagrubel:
> ./ci/integration_test.py -t checkpoint_restart,checkpoint_restart_failure

JFI: Didn't work until I loaded the charliecloud module, so if we make a document for running this remember to include that.

pagrubel:
This looks good, thank you!
@jtronge Could you put some instructions in the Readme for running the test, including ensuring charliecloud is loaded and  notes about cleaning up. I see there is a new directory .beeflow-integration 
Also, I'm wondering if we should have a special directory for workflows that are used for integration tests. the data/cwl/bee_workflows directory has some miscelleneous workflows in it that we may want to cleanup. Putting the ones used for testing in another directory would help us know they are used for that. What do you think?

jtronge:
I added a section in the developer docs with instructions for running tests locally.

I'm thinking I might put the test workflows in something like `ci/test_workflows`, removing them from `beeflow/data`. @pagrubel Do you think that makes sense?

pagrubel:
> I added a section in the developer docs with instructions for running tests locally.
> 
> I'm thinking I might put the test workflows in something like `ci/test_workflows`, removing them from `beeflow/data`. @pagrubel Do you think that makes sense?

Yes, that makes sense. Are they used for anything else?

jtronge:
No, it doesn't look like it. I did a `git grep` for the workflow directory names and it didn't show any other uses.

pagrubel (approved):
Looks good to me. I just made minor changes to the documentation so it shows up in the contents portion.

PR #770, opened 2024-01-17
Fix single CWL file parsing

### LINKED ISSUE #643: Single file CWL broken input

Boogie3D:
- The `cwl_utils` parser for some reason assigns slightly different step input ID values depending on whether the step is parsed from a separate CommandLineTool CWL file or directly from the `run` field in the Workflow CWL file `steps` section
  - e.g. separate file: `file://foo/bar#iterations`; inline run field: `file://foo/bar#lulesh/iterations`
- This caused a key lookup in the input-to-source map to fail, raising an exception
- Fixed by stripping the extraneous prefix (e.g. `lulesh/`) from the step input ID
- Fixed a related issue where the `TASKS_NOJOB_GOLD` tasks in `test_parser.py` had empty `inputs` lists, causing `test_parse_workflow_no_job()` to fail erroneously after the bug was fixed (it was passing before the fix)
  - The gold standard tasks had empty `inputs` lists because that's what the parser was returning, which is possibly another bug
  - It may need to be further investigated why the parser was returning tasks with empty `inputs` lists when parsing the `cf.cwl` test file rather than raising an exception like the lulesh-mpi workflow

This PR resolves #643


jtronge (approved):
This looks good. I tried the lulesh-mpi workflow and it worked.

rstyd:
I'll merge and close this. Thanks, Steven!

Issue #750, opened 2023-12-04
Checkpoint/Restart is not working.


pagrubel:
Using the workflow in beeflow/data/cwl/bee_workflows/clamr-wf-checkpoint that worked previously. Querying the status of the workflow gives:
```beeflow query 17
Running
clamr--TIMEOUT
ffmpeg--WAITING
```

The TIMEOUT state should spawn restart tasks.

pagrubel:
These are the entries at the end of the task manager log:
```[2023-12-04 17:22:44,426] beeflow.task_manager:submit_jobs(): Job Submitted clamr: job_id: 15569181 job_state: PENDING
[2023-12-04 17:24:09,374] beeflow.task_manager:update_jobs(): state: TIMEOUT
[2023-12-04 17:24:09,375] beeflow.task_manager:update_jobs(): TIMELIMIT/TIMEOUT task_checkpoint: {'file_path': 'checkpoint_output', 'num_tries': '3', 'container_path': 'checkpoint_output', 'enabled': 'True', 'restart_parameters': '-R', 'file_regex': 'backup[0-9]*.crx'}
[2023-12-04 17:24:09,544] beeflow.task_manager:update_task_state(): WFM not responding when sending task update.```

aquan9:
Steven fixed the retries type issue. It sounds like we need to go to wf_update.py.  The workflows are getting stuck in the timeout state when they run out of retries, but they should move to the "fail" state. Currently the branch he has to fix the issue is working to do the restart, but it just gets stuck on this last state update.

pagrubel:
Addressed in #776

PR #776, opened 2024-02-07
Fix Checkpoint/Restart

jtronge:
Steven's fixes for checkpoint/restart, addressing issue #750.

After multiple restarts of a task that exceeds `num_tries`, the workflow should be put in a `Failed` state. I'm not sure if this should be the `Archived` state instead, which I think happens with other workflows?

The checkpoint-restart tests can be run locally, after starting beeflow, with `./ci/integration_test.py -t checkpoint_restart_failure,checkpoint_restart --timeout 10000` (this might take >10 minutes).


pagrubel (approved):
In addition to the test's now passing, I ran the long checkpoint/restart test for clamr which failed since the max number of retries was set at 3 in my bee.conf.  The only concern I have is that the only indication as to what happened is in the wf_manager log. We definitely need something in the query to inform us about what happened, bu t that would be a separate issue. Also, the integration tests also failed for me due to an issue with not having any anonymous pulls left, again the issue can be found by examining the logs.
Thanks for fixing the checkpoint/restart  issue @Boogie3D & @jtronge

PR #778, opened 2024-02-08
Store submit command args in designated workflow directory.

### LINKED ISSUE #742: Archive should contain submit command
We need to know what files were used for submission especially if the submit command included an alternate yml or main_cwl file. 
Also, if the directory or tarball used for submission has several different options, we need to know what was used. This should be part of provenance but for now at least record the command.

kchilleri:
Addresses issue #742.


pagrubel (commented):
You might consider using fstrings 
f.write(f"wf_name:  {wf_name}\n")

pagrubel (approved):
Thank you, @kchilleri  you are now officially a contributor to the BEE project.

Issue #757, opened 2023-12-13
Reset: when a workflow's state is either Running or Waiting causes an error


pagrubel:
``beeflow core reset`` causes an error when workflows submitted but not started and when a workflow is running with the following error:
``OSError: [Errno 39] Directory not empty: 'x86_64-linux-gnu'``

This is caused by the neo4j instance still running. We can either check for ``Running`` or ``Waiting`` workflows, inform the user that they will need to cancel those workflows and not allow the operation or we can automatically cancel the workflows and continue the reset.

I like the first option.

pagrubel:
This is the current (in develop) result of trying this when there is either an Initializing or Running workflow.
```
beeflow core reset
A reset will remove this directory: /vast/home/pagrubel/.beeflow

Are you sure you want to reset?

Please ensure all workflows are complete before running a reset
Check the status of workflows by running 'beeflow list'

A reset will shutdown beeflow and its components.

A reset will delete the bee_workdir directory which results in:
Removing the archive of workflows executed.
Removing the archive of workflow containers.
Reset all databases associated with the beeflow app.
Removing all beeflow logs.

Beeflow configuration files from bee_cfg will remain.

Respond with yes(y)/no(n):  y
Beeflow has been shutdown.
Waiting for components to cleanly stop.
Traceback (most recent call last):

  File "/vast/home/pagrubel/.cache/pypoetry/virtualenvs/hpc-beeflow-YDRVf3zF-py3.11/bin/beeflow", line 6, in <module>
    sys.exit(main())
             ^^^^^^

  File "/vast/home/pagrubel/BEE/BEE/beeflow/client/bee_client.py", line 610, in main
    app()

  File "/vast/home/pagrubel/.cache/pypoetry/virtualenvs/hpc-beeflow-YDRVf3zF-py3.11/lib/python3.11/site-packages/typer/main.py", line 289, in __call__

  File "/vast/home/pagrubel/.cache/pypoetry/virtualenvs/hpc-beeflow-YDRVf3zF-py3.11/lib/python3.11/site-packages/typer/main.py", line 280, in __call__

  File "/vast/home/pagrubel/.cache/pypoetry/virtualenvs/hpc-beeflow-YDRVf3zF-py3.11/lib/python3.11/site-packages/click/core.py", line 1157, in __call__

[...]

PR #779, opened 2024-02-09
Issue757/fix reset error

pagrubel:
This PR should address issue #757 
In this PR a method to get a workflow list was added. This consolidated some code that was previously repeated in bee_client.py and that gave the ability to use the method in core.py to check for Running or Initializing workflows. If there are Initializing workflows a warning message is printed. I need to check if it should just fail, but I had a problem where I could not get rid of the intializing workflow and reset was needed. I'll test this a bit more before we should merge it.
If there are Running workflows the client exits with the recommendation of cancelling those workflows.
Later we could just cancel them automatically, but this PR will prevent errors.


pagrubel:
I based this off an unmerged branch so will close it and fix that

PR #785, opened 2024-02-15
CI Refactor

jtronge:
This is a small refactor of the CI scripts to allow for running jobs with other runners.

Also, beeflow will now check for `BEE_CONFIG` in your environment and use that path for the config file. Otherwise it will fall back to `~/.config/beeflow/bee.conf`.


pagrubel:
This looks good. I do have one question (not a stop on this PR). Going through a code review on this brought up the following question. We are behind on using the latest release of Charliecloud,  is there a way to find the latest and use it instead of hardcoding it?

pagrubel (approved):

Issue #675, opened 2023-04-10
Task Manager Resiliency: Check the status of any tasks that were running via slurm. Resubmit any failed tasks.


aquan9:
Pieces broken up from #614

pagrubel:
Should this be optional or automatic?
Should there be a configuration option that only tries x number of times?

jtronge:
I wonder if checkpoint-restart would affect this? We might want to only resubmit tasks if the `num_tries` in `beeflow:CheckpointRequirement` allows it. I'm not sure if the number of times a task has already been restarted is stored in a database, so we might lose it on task manager failure.

pagrubel:
I think this issue needs some discussion and clarification, maybe during a meeting.

jtronge:
Resolved with PR #827

Issue #676, opened 2023-04-10
Task Manager Resiliency: Verify that the workflow manager is back up before restarting tasks.


aquan9:
Pieces broken up from #614

aquan9:
Pieces broken up from #614

jtronge:
Resolved with PR #827

PR #789, opened 2024-02-28
Initial task manager resiliency and error handling

### LINKED ISSUE #550: Builder Errors
When the builder encounters certain errors, such as a Dockerfile not existing, etc., it will simply print a warning and the workflow will continue. This makes it difficult to figure out why a workflow isn't working and the builder's error message gets buried in the log. For example I tried running the `examples/clamr-ffmpeg-build` workflow and the builder printed a warning `[01;33mWARNING: There was a problem executing dockerFile![0m`. The workflow continued and just didn't produce any output files.

This is slightly related to issue #485, but I thought it was best to create a new one.

jtronge:
These are some initial changes to improve resiliency and error-handling in the task manager. This doesn't completely resolve issues #675 and #676, but I thought I'd open this now since these changes are relatively self-contained. Rusty and I discussed those issues and we were thinking it might be best to have a longer discussion about them and the interaction between the task manager and the workflow manager.

Also, I think this should resolve #550. The builder should now be throwing exceptions when it fails to build or pull a container. I added an integration test case for an invalid container build.

### DIFF
--- a/beeflow/client/bee_client.py
+++ b/beeflow/client/bee_client.py
@@ -459,11 +459,9 @@ def query(wf_id: str = typer.Argument(..., callback=match_short_id)):
 
     tasks_status = resp.json()['tasks_status']
     wf_status = resp.json()['wf_status']
-    if tasks_status == 'Unavailable':
-        typer.echo(wf_status)
-    else:
-        typer.echo(wf_status)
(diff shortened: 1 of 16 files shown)


aquan9 (approved):
This all looks good to me. Good job writing a new test for your changes.

PR #793, opened 2024-03-07
Fix reset error: addresses issue #757

pagrubel:
- check workflow state
    - do not allow reset when there are Running or Intializing workflows
    - add get_workflow_list method to eliminate duplicate code
    - make response text message clearer to read in code

Resolves: issue #757

PR #794, opened 2024-03-19
Fix reset error issue #757:  caused errors when there were active wor…

### LINKED ISSUE #757: Reset: when a workflow's state is either Running or Waiting causes an error

pagrubel:
…kflows. This commit performs the following to solve the errors:

 - allow reset when there are active workflows, remove workflows
 - add ability to remove or cancel paused workflows
 - check workflows for active states
 - warn user there are active states before continuing
 - stop all processes for active workflows
 - add get_workflow_list method to eliminiate duplicate code
 - make response text message clearer to read in code
 - add timestamp to name for unique backup if archive flag is set
 - simplify complexity of reset function
 - print warning and possible course of action when intializing workflows cause directory removal to fail

Resolves: #757

### DIFF
--- a/beeflow/client/bee_client.py
+++ b/beeflow/client/bee_client.py
@@ -106,15 +106,25 @@ def _resource(tag=""):
     return _url() + str(tag)
 
 
+def get_wf_list():
+    """Get the list of all workflows."""
+    try:
+        conn = _wfm_conn()
+        resp = conn.get(_url(), timeout=60)
+    except requests.exceptions.ConnectionError:
+        error_exit('Could not reach WF Manager.')
+
+    if resp.status_code != requests.codes.okay:  # pylint: disable=no-member
+        error_exit('WF Manager did not return workflow list')
+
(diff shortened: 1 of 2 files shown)


pagrubel:
This PR should be ready to merge. There is some discussion in #793 (closed) if you want to follow why this was implemented by removing all workflows. I could not remove initializing workflows, but the bee_workdir can be removed manually by the user once the reset is complete. Since this is mostly for developers I just added a message to tell them to do that.

kchilleri (approved):
This looks great! Also takes care of _initializing_ workflows issue I kept having.

Issue #795, opened 2024-03-19
Updating config file automatically


pagrubel:
In the past we've had problems when changing from one branch to another when a new config value has been added. This at times mean we need to rerun `bee config new` and saving the old then editting any values that we had put it there so that it matches the new branch. If we could come up with a way just to update the config file if a new value is needed and make it backwards compatible. That would make it easier to upgrade to a new version of BEE.

pagrubel:
This was fixed in PR #798

PR #798, opened 2024-03-20
Update config code to handle invalid/missing options

jtronge:
I think this should address issue #795. This updates the config validation code to display warnings for most invalid or missing options, instead of raising an exception. Options without defaults are still required and will cause validation to fail (like for `DEFAULT::workload_scheduler`).

I'm opening the PR now to run it through CI and hopefully catch remaining bugs.

### DIFF
--- a/beeflow/common/config_driver.py
+++ b/beeflow/common/config_driver.py
@@ -230,45 +230,41 @@ USER = getpass.getuser()
 VALIDATOR = ConfigValidator('BEE configuration file and validation information.')
 VALIDATOR.section('DEFAULT', info='Default bee.conf configuration section.')
 VALIDATOR.option('DEFAULT', 'bee_workdir', info='main BEE workdir',
-                 attrs={'default': DEFAULT_BEE_WORKDIR}, validator=validation.make_dir)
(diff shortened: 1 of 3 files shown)


pagrubel:
@jtronge The warning is not printing out correctly:
I deleted max_restarts from the config and got this. 

```
WARNING:beeflow.common.config_validator:option {sec_name}::{opt_name} is missing, setting to default "3"; please check that that is correct and update the config
```

jtronge:
Thanks @pagrubel. I think this last commit should have fixed that.

pagrubel:
I'll defer to Andres to review this to see if it helps with his branch.

aquan9 (approved):
This seems to work as far as I can tell. 

I tested it on the config file generated from my REST API branch. 

I get the following:
```
(hpc-beeflow-py3.9) [aquan@darwin-fe1 BEE]$ beeflow config validate
[2024-04-30 09:25:10,493] beeflow.common.config_validator:_validate_existing_options(): option DEFAULT::bee_droppoint is not a valid option
WARNING:beeflow.common.config_validator:option DEFAULT::bee_droppoint is not a valid option
```

Which I think is an appropriate warning.

PR #803, opened 2024-03-26
Fix CRT workdir bind mount

rstyd:
This PR fixes an issue that occurs on certain systems with symlinked home directories. Instead of bind mounting, the entire home directory, we just bind mount the workdir directly to the `/mnt/9/` directory in the container.


jtronge:
This looks good Rusty. I'll approve once the pylama error is fixed. I think you can just remove `beeflow/common/container_path.py` altogether.

rstyd:
Closed since it doesn't seem to be an issue on important systems.

PR #805, opened 2024-03-26
Fix minor error when attempting reset with no active workflows

pagrubel:
This PR fixes an error preventing `beeflow core reset` from working when there are workflows but none are active.


kchilleri (approved):
This works when there are only archived workflows

Issue #688, opened 2023-05-09
If the state of a step is FAILED, the steps that are dependent on it continue to run.


pagrubel:
If the state of a step is FAILED, the steps that are dependent on it continue to run.
You can try this using the cat-grep-tar example by running in a workdir that does not contain lorem.txt.
Here is my submit command:
```
 beeclient submit cgtnolorem ~/BEE/BEE/examples/cat-grep-tar ~/BEE/BEE/examples/cat-grep-tar/workflow.cwl ~/BEE/BEE/examples/cat-grep-tar/input.yml ~/beeworkdir/
```
And the resulting query:
```
beeclient query 1b6517
Archived
grep0--FAILED
cat--FAILED
grep1--FAILED
tar--COMPLETED
```
By default I think grep0, grep1 and tar should not run. We could make it an option. I cannot think of anytime a step that is dependent on a failed step should run though.

pagrubel:
Addressed in #806

PR #806, opened 2024-03-28
Ensure failed tasks cause dependent tasks to fail

jtronge:
This should address issue #688. When a task fails, all dependent tasks will get a state of `DEP_FAIL` and the workflow will fail without running anything else.

### DIFF
--- a/beeflow/common/integration/utils.py
+++ b/beeflow/common/integration/utils.py
@@ -248,5 +248,5 @@ def check_completed(workflow):
 
 def check_workflow_failed(workflow):
     """Ensure that the workflow completed in a Failed state."""
-    ci_assert(workflow.status == 'Failed',
+    ci_assert(workflow.status == 'Archived/Failed',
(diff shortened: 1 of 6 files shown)


pagrubel:
@jtronge I tried this out. It works well. Even though the workflow doesn't finish I think it should tar up the artifacts and archive them.

jtronge:
@pagrubel Should the workflow state remain `Failed` when an unfinished workflow gets archived?

rstyd:
How about ARCHIVED/Failed?

jtronge:
I think `Archived/Failed` should be good. Failed workflows should be getting archived now, including the ones caused by build failures.

We seem to be having poetry problems on recent CI runs. Poetry's failing with`Package docutils (0.21.post1) not found.`.

pagrubel:
Thanks @jtronge I like the Archive/Failed state and still have the archive for error analysis.

pagrubel (approved):

PR #808, opened 2024-03-30
Fix issues building the documentation

aquan9:
The error in building the last documentation, seems to stem from the issue documented here: https://github.com/readthedocs/readthedocs.org/issues/9038

I added a constraint as recommended by that thread to our pyproject.toml to fix that. 

I tested that fix in a virtual github codespace.


pagrubel:
Thank you Andres, I'm going to close this since we quit using jinja and should not have been referencing it. I'll open it back up if we need it.

PR #811, opened 2024-04-01
Fix beeflow config new error

jtronge:
This moves an indirect call to `bc.get(...)` from the beeflow code into a function to avoid errors when the config doesn't exist yet.


rstyd (approved):
That works!

Issue #817, opened 2024-04-09
Upgrade Neo4j version


jtronge:
We need to upgrade Neo4j to the 5.x version since our current version is deprecated.

pagrubel:
Addressed by #819

PR #819, opened 2024-04-11
Update Neo4j to 5.x

jtronge:
This should update BEE to use neo4j 5.x. See issue #817. I've updated the Python API and made a couple changes to how neo4j is launched.

This is still a WIP but I'm opening it now to see if the CI passes.


pagrubel:
@jtronge Is this still WIP? Maybe we need to revisit our labels.

jtronge:
I think this might be ready for more local testing now.

Maybe we should change WIP so that it always runs CI actions? That's the only reason I left it off, so that I could run this PR through CI. I think we set it that way in the first place to avoid wasting CI time, but I'm not sure that's still a major problem.

pagrubel (approved):
This looks good. I made minor changes to documentation and the feature to pull the container.

PR #820, opened 2024-04-15
Specify docutils version to fix poetry/docutils bug

jtronge:
See the poetry issue here: https://github.com/python-poetry/poetry/issues/9293.


pagrubel:
Looks good, thanks for track this down!

pagrubel (approved):

PR #824, opened 2024-04-17
Archive failed checkpoint restart workflows

jtronge:
Failed checkpoint restart workflows should now be archived with a state of `Archived/Failed`.

This should fix the nightly CI.


pagrubel (approved):
Thanks for fixing this so quickly Jake. It looks good and I tried it out with a checkpointing workflow that I made fail early.

PR #826, opened 2024-04-22
Fix status display

pagrubel:
Corrects the display in the documentation for ``beeflow core status``.  The code block was not displaying.


kchilleri (approved):
This works. The status block is being displayed now.

PR #827, opened 2024-04-25
Update TM for task update resiliency

jtronge:
I think this should address issues #676 and #675. This adds an update queue to store task updates from the Task Manager to ensure they aren't lost if the Workflow Manager is down.

I'm marking this WIP since we had the Task Manager resiliency meeting several weeks ago and I may have forgotten something here.


rstyd (approved):
This looks great. Thanks, Jake!

PR #828, opened 2024-04-25
Allow removal of a workflow with any Archive status

pagrubel:
When attempting to remove a workflow with the new Archive/Failed status, it was not allowed because beeflow wasn't aware of this status and thought the workflow might be running. This PR now allows any workflow with a status containing Archive to be removed.

### DIFF
--- a/beeflow/client/bee_client.py
+++ b/beeflow/client/bee_client.py
@@ -374,7 +374,7 @@ def remove(wf_id: str = typer.Argument(..., callback=match_short_id)):
 
     wf_status = get_wf_status(wf_id)
     print(f"Workflow Status is {wf_status}")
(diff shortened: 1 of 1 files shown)


rstyd:
This looks good to me! Approved.

rstyd (approved):

pagrubel:
I'd like to understand this a bit better in light of the work I'm doing concerning slurm, so putting the WIP label back on so we don't merge it yet.

rstyd:
Gotcha thanks!

pagrubel:
Sorry this was not the PR I wanted to hold off on. So I removed the WIP lablel

rstyd:
Oh okay in that case I'll merge it.

Issue #783, opened 2024-02-14
Pause active workflow for ``beeflow core stop`` 


pagrubel:
We should probably just check for running or initializing workflows and not allow a stop, instead ask the user to cancel such workflows. 

If the user goes ahead with ``beeflow core stop`` when there is a running workflow, it can leave an orphaned .nfs file and is hard to do a reset or other clean up activities.

pagrubel:
There is a PR that is similar for ``beeflow core reset`` so I suggest this be done once that is merged

pagrubel:
After testing it seems that the best thing to do would be to warn the user if there are running workflows and ask to proceed and if they do to pause the workflows and tell them they will need to resume them.
Testing this on darwin overnight does not work well. It loses track of workflows that were running when beeflow core stop happened.

pagrubel:
Closing this as it has been fixed

PR #830, opened 2024-04-29
Pause running workflows when stopping beeflow with 'beeflow core stop'

pagrubel:
This PR pauses any running or waiting workflows when stopping beeflow.
Tasks/jobs that are running or pending when beeflow is stopped will be updated, if possible, once beeflow starts up again.
Addresses #783

### DIFF
--- a/beeflow/task_manager/background.py
+++ b/beeflow/task_manager/background.py
@@ -30,7 +30,7 @@ def submit_task(db, worker, task):
         resolve_environment(task)
         log.info(f'Environment preparation complete for task {task.name}')
         job_id, job_state = worker.submit_task(task)
-        log.info(f'Job Submitted {task.name}: job_id: {job_id} job_state: {job_state}')
+        log.info(f"Job Submitted '{task.name}' job_id: {job_id} job_state: {job_state}")
         # place job in queue to monitor
         db.job_queue.push(task=task, job_id=job_id, job_state=job_state)
         # update_task_metadata(task.id, task_metadata)
@@ -75,10 +75,10 @@ def update_jobs(db):
         # If state changes update the WFM
         if job_state != new_job_state:
             db.job_queue.update_job_state(id_, new_job_state)
+            log.info(f"Job Updated '{task.name}' job_id: {job_id} job_state: {new_job_state}")
             if new_job_state in ('FAILED', 'TIMELIMIT', 'TIMEOUT'):
                 # Harvest lastest checkpoint file.
                 task_checkpoint = task.get_full_requirement('beeflow:CheckpointRequirement')
-                log.info(f'state: {new_job_state}')
                 log.info(f'TIMELIMIT/TIMEOUT task_checkpoint: {task_checkpoint}')
                 if task_checkpoint:
                     try:
@@ -94,7 +94,6 @@ def update_jobs(db):
             # States are based on https://slurm.schedmd.com/squeue.html#SECTION_JOB-STATE-CODES
             elif new_job_state in ('BOOT_FAIL', 'NODE_FAIL', 'OUT_OF_MEMORY', 'PREEMPTED'):
                 # Don't update wfm, just resubmit
-                log.info(f'Task {task.name} in state {new_job_state}')
                 log.info(f'Resubmitting task {task.name}')
                 db.job_queue.remove_by_id(id_)
                 job_state = submit_task(db, worker, task)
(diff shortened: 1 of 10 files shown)


pagrubel:
I need to do some overnight testing on this branch.

pagrubel:
To test this for CLI (on darwin) you may want two screens both in the poetry env for this branch:
  - Make sure use_commands in the slurm portion of bee.conf is True.
  - Start beeflow: beeflow core start
  - submit a workflow, clamr example works well, the checkpoint workflow is also good
      ``watch -n 2 query <wf_id>`` to make sure the task is pending or running. 
       It is helpful to do this in a separate screen and keep it up even when you stop beeflow, or when you restart it
  - verify that the clamr step is running and get the job id via ``squeue -u <username> `` or the task manager log
  - issue ``beeflow core stop`` while clamr step is running
  - ``watch -n 5 show job <job_id>`` Wait until this gives a no job type error
  - Then start beeflow up again ``beeflow core start``. The query screen will show that clamr has completed.
  - The workflow has been paused so to finish it just submit the  ``beeflow resume <wf_id>``

To test for slurmrestd:
- Make sure use_commands in the slurm portion of bee.conf is False.
  - Start beeflow: beeflow core start
  - submit a workflow, clamr example works well, the checkpoint workflow is also good
      ``watch -n 2 query <wf_id>`` to make sure the task is pending or running. 
       It is helpful to do this in a separate screen and keep it up even when you stop beeflow, or when you restart it
  - verify that the clamr step is running and get the job id via ``squeue -u <username> `` or the task manager log
  - issue ``beeflow core stop`` while clamr step is running
  - ``watch -n 5 squeue -u <userid>`` Wait until the clamr job is off the screen
  - Then start beeflow up again ``beeflow core start``. The query screen will show that clamr has completed.
  - The workflow has been paused so to finish it just submit the  ``beeflow resume <wf_id>``

pagrubel:
Adding this thought here to implement later: The output from "sacct -j <job_id>" is in column form where the job_id is a row heading and Job State is a column heading. That should be used instead of the location in case another column or row is added in the future.

pagrubel:
@jtronge What happens if sacct throws an exception with job_state? I was hoping to get it out of the queue but not sure it will be.

pagrubel:
@jtronge I tested this on the prod system and it works. The only additional feature we may want to add at some point is for the user to be able to resume all paused workflows with one command.

beeflow/task_manager/background.py
```diff
@@ -94,15 +94,14 @@ def update_jobs(db):
...
-                log.info(f'Task {task.name} in state {new_job_state}')
                 log.info(f'Resubmitting task {task.name}')
                 db.job_queue.remove_by_id(id_)
                 job_state = submit_task(db, worker, task)
                 db.update_queue.push(task.workflow_id, task.id, job_state)
             else:
                 db.update_queue.push(task.workflow_id, task.id, new_job_state)
 
-        if job_state in ('ZOMBIE', 'COMPLETED', 'CANCELLED', 'FAILED', 'TIMEOUT', 'TIMELIMIT'):
+        if job_state in ('UNKNOWN', 'COMPLETED', 'CANCELLED', 'FAILED', 'TIMEOUT', 'TIMELIMIT'):
```
jtronge, line 104: Does this get rid of `ZOMBIE` altogether?

jtronge (commented):
I think this looks good. Since I made a couple of commits, should someone else approve as well?

jtronge:
> @jtronge What happens if sacct throws an exception with job_state? I was hoping to get it out of the queue but not sure it will be.

Oh sorry, just saw your question. I think this depends on what the calling code is doing. The TM background code could try to catch `WorkerError` and remove the job in that case.

beeflow/task_manager/background.py
pagrubel: I think ZOMBIE was meant for Workflow states

pagrubel: Not for jobs

pagrubel:
> > @jtronge What happens if sacct throws an exception with job_state? I was hoping to get it out of the queue but not sure it will be.
> 
> Oh sorry, just saw your question. I think this depends on what the calling code is doing. The TM background code could try to catch `WorkerError` and remove the job in that case.

Should we put WIP back on do this. I'm not sure there is any way to test it.

aquan9 (approved):
This all looks ok to me.

PR #838, opened 2024-05-08
Add integration test for pre/post scripts (beeflow:ScriptRequirement)

jtronge:
This adds a test for the pre/post script requirement and also refactors some of the set up code in the integration tests.


rstyd (approved):
Looks good to me!

Issue #823, opened 2024-04-16
Add test (ci) when slurmrestd is not available


pagrubel:
We have a path for systems that do not have slurmrestd. We need to test it and make sure it doesn't break.

pagrubel:
Do we just want to set use_commands to True in the bee.conf file to test this.

jtronge:
We might be able to test this with the integration CI. We would need to update `.github/workflows/integration.yml` to also test a slurm+commands instance. Right now that workflow runs two instances of the integration tests, one for slurm+slurmrestd and one for flux.

aquan9:
This issue is complete for the Github actions CI. 

See the following in ci/bee_config.sh:
```
case $BEE_WORKER in
Slurmrestd)
    cat >> $BEE_CONFIG <<EOF
[slurm]
use_commands = False
openapi_version = $OPENAPI_VERSION
EOF
    ;;
SlurmCommands)
    cat >> $BEE_CONFIG <<EOF
[slurm]
use_commands = True
openapi_version = $OPENAPI_VERSION
EOF
    ;;
esac
```

PR #839, opened 2024-05-13
Test slurm commands worker in CI

jtronge:
This should add testing for the slurm commands version of the worker in CI, addressing #823.

I'm labeling this WIP until the tests pass.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -18,9 +18,9 @@ jobs:
     name: BEE Integration Test
     strategy:
       matrix:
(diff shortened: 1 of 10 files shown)


rstyd:
This looks great! I think we should hold off on merging this until the restart GDB pr is merged just cause I don't want to have to rebase it again. lol

rstyd (approved):

pagrubel:
Ugh this is the one I'd like to understand better before merging. I guess we should just hold off on merging anything right now.

PR #841, opened 2024-05-20
Generate CWL

rstyd:
This PR adds code to generate a bee compatible CWL workflow using python.


pagrubel:
@rstyd Do you have an example of a multi-step workflow?

aquan9:
Looks like it's failing linting at the moment.

aquan9 (commented):
One thing I noticed that's still missing is some documentation. The example is good. I'll give it a shot, but it would be good to also have a page in our documentation saying how to use it. Or maybe make an issue.

rstyd:
I'll make an issue, and we can just use it internally for now.

pagrubel:
I agree with Andres about documentation but am also fine with that being a separate PR. Having an issue for it will suffice.

pagrubel (approved):
We will need PR's for fixing the comd example and adding some documentation, however the generate code works for simple workflows and for cat-grep-tar with dependencies.

Issue #790, opened 2024-03-04
Invalid workflow step inputs are not caught until runtime


jtronge:
Invalid input names are not getting caught until the task manager attempts to construct the command on submission, resulting in a SUBMIT_FAIL for the task. We should probably try to catch these in the client at parse time.

pagrubel:
@jtronge Is this the name of the input file?

rstyd:
This was completed with #842.

PR #842, opened 2024-05-23
Check for missing workflow inputs

jtronge:
This updates the parser to check for missing workflow inputs (addressing issue #790).

I'm not sure if this is the right way to handle this so I'm marking this WIP for now.


rstyd (approved):
This works perfectly! Approved.

PR #844, opened 2024-05-28
validate shell environment in pre/post script files

### LINKED ISSUE #801: Add $SHELL option to pre/post script
The pre/post script requirement currently just uses whatever shell is defined in `$SHELL` for the user. We should add an additional parameter called `shell` which enables a user to specify what shell interpreter they want to be used for their scripts. A lot of systems use `tsch` or `csh` by default while many users actually use bash in practice so setting this in the workflow will avoid any conflicts between the users local configuration and the configuration of whoever created the workflow.

My proposed options are:

* sh
* bash
* csh 
* tcsh
* zsh

We could also just allow it to use any shell in `/etc/shells` on the machine, but that might make things over complicated.

kchilleri:
When pre/post script files are enabled, at parse time, it checks the files for a shebang line (first line of file). If no shell environment is defined, it returns an error.

### DIFF
--- a/beeflow/common/parser/parser.py
+++ b/beeflow/common/parser/parser.py
@@ -282,7 +282,7 @@ class CwlParser:
 
     def _read_requirement_file(self, key, items, state=0):
         """Read in a requirement file and replace it in the parsed items.
-        
+
         :param state: tracks if pre/post script capability is enabled
         :type state: 0 is default, 1 is True
         """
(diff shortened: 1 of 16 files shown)


kchilleri:
To Do:
1. Return error if no shebang line exists: what kind of error do we want to use here? Currently, I have a print statement stating no shebang line was found.
2. Modify documentation.

rstyd (approved):
Approved!

PR #852, opened 2024-06-11
Security concrete

aquan9:
This PR adds in the poetry.lock file which should hopefully fix the issues we've been having with dependabot not being able to automatically generate PRs for detected issues. 

This PR also starts from the Gitlab CI coverage badge fix as a base so that tests pass. Hence the larger commit history. 

See below screenshot:

Screenshot 2024-06-10 at 11 26 34 PM


rstyd (approved):
Looks good to me!

Issue #845, opened 2024-05-29
Add $SHELL Option to BEE ScriptRequirement


rstyd:
We're currently making the unspoken assumption that BEE users will always prefer to use BASH for their `sbatch` scripts. That hasn't been an issue previously, but with the addition of the pre/post script support we have to ensure that the code within those scripts is interoperable with the `sbatch` script they will be embedded within. 

For this issue, we should add an additional option to the `beeflow:ScriptRequirement` that specifies which `$SHELL` the user prefers. I think this option should be the path to the `$SHELL` e.g. `/bin/bash`, `/bin/csh/`, `/bin/tcsh`.

jtronge:
I'm thinking we might want to have this per-workflow rather than in the config file. I can see where some workflows would have to use csh for some code while bash for another one, as an example. Maybe it could be a part of `beeflow:ScriptRequirement`?

pagrubel:
That sounds good. We could have default defined in the config though so that those who don't change do not have to specify it.

kchilleri:
@rstyd have you pushed your code for this issue anywhere?

kchilleri:
PR#856 closes this issue

PR #856, opened 2024-06-18
Add Script Option to Pre/Post Script & Change MPI Call

rstyd:
This PR addresses #845. It makes `/bin/bash` the default shell and changes the approach we use for MPI.


jtronge (approved):
Looks good. I tested the comd-mpi workflow with this and it works.

kchilleri (approved):
Tested the new beeflow:ScriptRequirement shell option, this works for me!

Issue #857, opened 2024-06-18
Add integration test for MPI jobs


jtronge:
We should add an integration test for MPI jobs. Right now we can only test this manually with workflows in `beeflow/data`.

pagrubel:
Addressed in #858

PR #858, opened 2024-06-20
Add MPI integration test code

jtronge:
This should address issue #857. I added a test using the comd-mpi workflow. To run this, first start beeflow core and then run `./ci/integration_test.py -t comd_mpi --timeout 10000`, which should result in `PASSED`.


pagrubel:
> This should address issue #857. I added a test using the comd-mpi workflow. To run this, first start beeflow core and then run `./ci/integration_test.py -t comd_mpi --timeout 10000`, which should result in `PASSED`.
@jtronge maybe add this as a README

jtronge:
Should I update this in the developer docs? See https://lanl.github.io/BEE/development.html#running-tests. I guess there's not too much info there about running the integration tests.

pagrubel:
Sure

pagrubel:
> Should I update this in the developer docs? See https://lanl.github.io/BEE/development.html#running-tests. I guess there's not too much info there about running the integration tests.

I guess that could be another issue.

pagrubel:
This works for me running on 4 nodes, thanks!

pagrubel (approved):

Issue #835, opened 2024-04-30
Should we disable bee enhanced client


pagrubel:
There is a security alert for lodash only used for the enhanced client. 
We haven't been using bee enhanced client so disabling should be okay.

aquan9:
While discussing this in a meeting, Jake also agrees that this should be disabled. We haven't used it in over a year. It hasn't been put into use, and it just collects security vulnerabilities.

pagrubel:
Addressed in #865

PR #865, opened 2024-06-24
Remove enhanced client from github

rstyd:
This PR addresses #835.


aquan9 (approved):
This looks good to me. The enhanced client is deprecated, and continues to just cause security issues and is not under active development. A copy has been stashed on the re-git Gitlab instance under the BEEdev group.

PR #879, opened 2024-07-12
Issue871/delete_use_archive_config

### LINKED ISSUE #871: use_archive in the configuration file bee.conf is currently unused
The purpose of use_archive is to signal whether or not workflows should be archived. The default is True.
 behavior should be

If use_archive is True then all workflows are archived.
If use_archive is False then no workflows are archived. 
              
However, this value is currently not even being used. 

Do we want to have the ability to turn off archiving? or 
Should we just remove the configuration parameter?

Leahh02:
Removed use_archive from config_driver.py and bee_config.sh

I couldn't find anything in the documentation that referenced use_archive so I didn't update any documentation.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -56,6 +56,7 @@ Contributors:
 * Patricia Grubel - `pagrubel <https://github.com/pagrubel>`_
 * Qiang Guan - `guanxyz <https://github.com/guanxyz>`_
 * Ragini Gupta - `raginigupta6 <https://github.com/raginigupta6>`_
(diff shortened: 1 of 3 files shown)


pagrubel (changes requested):
@Leahh02 use_archive remains in the config file. There is some code in beeflow/common/config_driver.py  for the validator that needs to be removed. see line 236-7

pagrubel (approved):

pagrubel:
This looks good and satisfies #871 
Thank you!

PR #880, opened 2024-07-16
Validate shell option in cwl file and make sure it matches shell interpreter in pre/post scripts

### LINKED ISSUE #847: Pre/Post Scripts should use the same $SHELL as defined in the BEE config
This is a follow on to #845 and #846. Once we have a proper BEE shell configuration option and the sbatch scripts follow that, we'll need to ensure that the pre/post scripts also follow the `$SHELL` config.

@kchilleri is currently working on verifying that pre/post scripts have a shebang that defines _a_ `$SHELL` interpreter at parse time. We could additionally check after that the interpreter defined within that shebang matches the` $SHELL` defined in the BEE config and throw an error if it doesn't.

kchilleri:
This PR closes issue #847.

It makes sure the shell option is defined in the cwl file if pre/post scripts are enabled (it returns an error if not).

Then it checks that the shell interpreter defined in the cwl file matches the shell interpreter defined in the pre/post scripts (it returns an error if not).

This PR also adds error checking and testing that was not added in PR #856.

CI integration test "pre_post_script()" will test this added code so there is no need to add another one.

Documentation is modified accordingly.


rstyd (approved):
Great job! I approve.

PR #885, opened 2024-07-23
added check for /tmp or /var/tmp

Leahh02:
Checks if the workdir is /tmp or /var/tmp and send an error_exit if it is. Takes into account when /tmp or /var/tmp are symbolic links pointing to another directory. For example my /tmp points to /ram/tmp and my /var/tmp points to /ram/var/tmp. I wasn't sure if the symbolic links would be the same on other clusters, so I made it general.


rstyd (changes requested):
I spoke with Leah offline. This is _almost_ perfect, but we want to verify that the workdir isn't in `/tmp` or `/var/tmp`.

rstyd (approved):
This works perfectly! Approved!

Issue #859, opened 2024-06-20
Update test directions in developer manual


pagrubel:
See https://lanl.github.io/BEE/development.html#running-tests. There's not too much info there about running the integration tests.

pagrubel:
This was addressed in #887

PR #887, opened 2024-07-24
Update documentation for the integration tests

jtronge:
I think this should address issue #859, at least for the integration tests. I've added some more info about what tests are actually being run, how to show those, and how to run tests not on by default.


pagrubel (commented):
@jtronge This looks good. I don't know why but when I build it the reference to the command line interface doesn't work correctly, but it doesn't work correctly for me on develop branch either.

pagrubel (changes requested):
I found the reference problem, there needs to be two dots in commands.rst at the beginning of line 1 not one dot.

rstyd (approved):
Looks good to me!

rstyd (approved):
This addresses the requested changes.

pagrubel (approved):

Issue #874, opened 2024-07-05
Checking job state after it already has a COMPLETED (or other state that says its done) sometimes giving "NOT_RESPONDING"


pagrubel:
It seems we may have a race condition where the job has completed, failed or timed out and we still check for it. There have been a few instances where the state has changed from one of those states, COMPLETED to NOT_RESPONDING however the workflow continues because the dependency was set as satisfied. A fix for this is to not ever update a state if it has been one of the states that indicate it is done.

pagrubel:
This needs to be done both in flux and slurm workers

pagrubel:
Completed states are: FAILED, TIMEOUT, COMPLETED. check what happens if a job is cancelled. If any of these states are reached we should not check for new states. Occurs on fast systems.

pagrubel:
Satisfied in #892

PR #892, opened 2024-07-29
Stop checking job state when it's completed

Leahh02:
I tried to replicate the "NOT_RESPONDING" issue before working this, but I wasn't able to. Let me know if this fix should be sufficient to solve the problem. Also, let me know if anything should be added to the COMPLETED_STATES set. It's currently "COMPLETED_STATES = {"FAILED", "TIMEOUT", "COMPLETED"}".


pagrubel:
@Leahh02 I haven't reproduced the NOT_RESPONDING state, but I do think you need to add CANCELLED to that list of COMPLETED_STATES

beeflow/task_manager/background.py
```diff
@@ -70,6 +72,11 @@ def update_jobs(db):
         task = job.task
         job_id = job.job_id
         job_state = job.job_state
+
+        if job_state in COMPLETED_STATES:
+            db.job_queue.remove_by_id(id_)
+            continue
```
pagrubel, line 80: I don't understand the use of continue here, if you remove the task from the queue should it check for state updates?

Leahh02: My thought process was that if a task is completed, it should be removed from the queue and then the "for job in job_q" loop should move onto the next job. The continue skips the check state updates and moves onto the next job. I was hoping this could fix the race condition because if a job is in a completed state it doesn't check for updates anymore.

Leahh02:
When I tried to add my commit "45acaa7" the unit tests failed. I think it was because there was stuff added into develop and I forgot to integrate those changes into my code. So I added Krishna's and Jake's commits to my branch. If I shouldn't have done that let me know.

pagrubel:
@Leahh02 Thank you this looks great. I'm just running a few of the tests we run on our overnight suite before approving.

pagrubel (approved):

pagrubel:
Addresses issue #874

PR #895, opened 2024-08-01
Remove shebang lines from pre/post scripts

### LINKED ISSUE #890: Remove Shebang from Pre and Post scripts After Validation
We need to remove the shebang part from the pre and post scripts since it interferes with the shebang at the beginning of the script. We also should ensure that it matches the SHELL option, but I'll make a separate issue for that since this is currently higher priority.

kchilleri:
This PR closes issue #890.

First, shebang lines from pre/post scripts are identified. Then, they are validated against the ```shell``` interpreted defined in the ```beeflow:ScriptRequirements``` section. After this validation, we remove the shebang lines from the pre/post scripts to avoid duplicates in the final running script.

A unit test (```def test_parse_workflow_script()``` has been modified to test this code.


rstyd (approved):
Approved!

PR #903, opened 2024-08-20
Visualize dag

Leahh02:
This branch is one that comes off of Kabir's Multiple-Workflows-One-Database branch. The purpose of it is to enable users to look at the dag without having to go to the web browser. I want to save the dag in a common format like .gml or dot so then users can put that file into a tool like graphviz. 

Before I made the pull request I did ``git merge develop`` because the Multiple-Workflows-One-Database branch didn't have changes from develop.


Leahh02:
Currently in wf_utils I'm triggering exporting of the dag within the --no-start flag. I'm planning on making a new flag, I just wanted to see how things worked first.

beeflow/client/bee_client.py
Leahh02, line 1: The changes made to bee_client.py and wf_utils.py are causing the NOT_RESPONDING state to appear -- I think. But this is my first attempt at having the "beeflow dag" command.

pagrubel: Do the steps actually complete and then show NOT_RESPONDING ?

Leahh02: Yes

generate_graph.py
Leahh02, line 1: Just to note, I'm not planning on keeping this here. I just wanted to put my work somewhere while I'm figuring things out.

Leahh02:
Switched to PR Export dag #909

PR #908, opened 2024-08-28
beeflow core info

aquan9:
This PR adds a way to get information about the current instance, which should be useful for debugging various issues:

Usage:
```
(poetry-ci) [aquan@darwin-fe1 .beeflow]$ beeflow core info
Beeflow version: 0.1.9.dev3
.beeflow directory: /vast/home/aquan/.beeflow
Log path: /vast/home/aquan/.beeflow/logs
```

### DIFF
--- a/beeflow/client/core.py
+++ b/beeflow/client/core.py
@@ -425,6 +425,13 @@ def status():
     for comp, stat in resp['components'].items():
         print(f'{comp} ... {stat}')
 
+@app.command()
+def info():
+    """Get information about beeflow's installation."""
+    version = importlib.metadata.version("hpc-beeflow")
+    print(f"Beeflow version: {version}")
+    print(f".beeflow directory: {paths.workdir()}") 
+    print(f"Log path: {paths.log_path()}")
 
 @app.command()
 def stop(query='yes'):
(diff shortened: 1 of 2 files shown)


pagrubel (changes requested):
@aquan9 I think it would be a good idea to add the container_archive path and the path's to our dependency containers.

pagrubel:
@aquan9 Also instead of .beeflow directory it should probably be with either "bee_workdir"

beeflow/client/core.py
```diff
@@ -426,6 +426,15 @@ def status():
         print(f'{comp} ... {stat}')
 
 
+@app.command()
+def info():
+    """Get information about beeflow's installation."""
+    version = importlib.metadata.version("hpc-beeflow")
+    print(f"Beeflow version: {version}")
+    print(f".beeflow directory: {paths.workdir()}")
```
pagrubel:
```suggestion
    print(f"bee_workflow directory: {paths.workdir()}")
```

pagrubel (changes requested):

pagrubel (approved):

pagrubel (approved):

PR #909, opened 2024-08-28
Export dag

Leahh02:
This is my work from the visualize-day branch, but it's now coming off of develop.

The goal is to export a dag as a graphml so it can be used by visualization tool.


Leahh02:
The changes in commit 2e284d15d945120efbe08fdd7cff140a104a325a is what made the "NOT_RESPONDING" state appear the other day. I'll test this again.

Leahh02:
I tested it again and the "NOT_RESPONDING" state didn't appear.

pagrubel:
Thank you this is much easier to review.

generate_graph.py
```diff
@@ -0,0 +1,56 @@
...
+        else:
+            dot.edge(target, source, label=edge_label, fontsize="10")
+
+    # Set the output format to PNG and render the graph
+    dot.format = 'png'
+    dot.render(output_path, view=False)
+
+if __name__ == "__main__":
+    # Path to your GraphML file
+    graphml_file = '/vast/home/leahh/.beeflow/dags/bebd6f.graphml'
```
aquan9: Please remove this user specific path, and replace it with a generic path for this PR.

generate_graph.py
aquan9, line 1: I think it would also be better if we move this file inside of the gdb directory, inside of common. 

In otherwords .../beeflow/common/gdb/generate_graph.py

aquan9 (dismissed):
This is great work, thanks Leah.  

One last thing I'd like to see (besides the other changes I noted) before putting my final stamp of approval is an addition to our beeflow client documentation, since this change adds a new command to the client.

pagrubel:
@Leahh02 uses an augmented neo4j container for this capability. I'd like to see the Dockerfile for it in our repo to have a record of what to build. We will also need to fix the dependency build to use that, but that can be done in another PR.
I was able to verify that merging this would not affect running with the basic neo4j container. The dag functionality won't work but workflows will.

Leahh02:
The clamr graph doesn't work, I'm trying to figure out why. I forgot to test it.

Leahh02:
> The clamr graph doesn't work, I'm trying to figure out why. I forgot to test it.

I can create a dag when I have submitted the workflow with the no-start flag, but once I've started the workflow I can't make the graph anymore. The graphml file gets created but there's problems with networkx.

pagrubel (commented):
This looks great! I made sure Andres's change requests were in this PR. I tested this and will run our local overnight tests, upon those passing I'll approve and merge.

pagrubel (approved):
Great job, thank you @Leahh02

PR #910, opened 2024-08-28
Updating pre-release to version 0.1.9.dev4

pagrubel:
Updating so we can make a pre-release that uses only on Neo4j instance.


rstyd (approved):
LGTM!

PR #914, opened 2024-09-03
No Start State

Leahh02:
This should allow the user to cancel a workflow submitted with the no-start state. The correct state should also come up when they query the workflow. The correct state should also appear in the graph database.


Leahh02:
I should probably standardize the capitalization. For example, when I query the workflow I get 
```
No Start
cat--No Start
grep0--No Start
grep1--No Start
tar--No Start
```
and when I start and then query I get 
```
Running
cat--COMPLETING
grep0--WAITING
grep1--WAITING
tar--WAITING
```

So I need to make it NO START

pagrubel:
I would think the Workflow state would be Not Started and the tasks could be blank.
Waiting for Submit for the Workflow

Leahh02:
> I would think the Workflow state would be Not Started and the tasks could be blank. Waiting for Submit for the Workflow

I can do that instead!

Leahh02:
After commit 8ba84cf6342d0678781279fc129366b4762344d1 the output looked like:
```
No Start
cat--
grep0--
grep1--
tar--
```
Then in commit 9841e8f731944cacb41585d86ba3c092bba77ff9 I decided to remove the -- so now it looks like:
```
No Start
cat
grep0
grep1
tar
```

pagrubel:
This looks good to me and works fine.

pagrubel:
@Leahh02 This looks good to me and works well. You will need to fix the conflicts. I've asked @kabir-vats to review it, if he has time.

pagrubel (approved):
Nice addition to allow a user to specify a workflow and verify the tasks and a first step in verifying the workflow dependencies.

PR #915, opened 2024-09-05
Change default test target for Slurm to be slurm commands

aquan9:
Trying to change the default internal test target to use slurm commands rather than slurmrestd to aid in testing.


rstyd (approved):
LGTM!

Issue #919, opened 2024-09-10
Workflow Status Updated in Multiple Places


kabir-vats:
The workflow status is updated in multiple places, including the workflow manager database, graph database and text file in the workflow archive. To sync these databases, methods should be refactored such that every change to workflow status calls one method.

PR #920, opened 2024-09-10
Refactor workflow status

kabir-vats:
Addresses #919 
The goal of this PR is to only change workflow status through one method (wf_utils.update_wf_status()) to ensure workflow status are synced across all databases.


kabir-vats:
Closing because outdated

PR #922, opened 2024-09-12
Reorganize common and wf manager

kabir-vats:
This change should make the common branch more easy to navigate and delete some ancient files that are no longer in use.


kabir-vats:
Closing because outdated

PR #923, opened 2024-09-13
Add task state when creating task for checkpoint

pagrubel:


pagrubel:
I was able to trigger the internal test for this on gitlab, that tests checkpoint restart and it passed, so it is ready to merge.

Leahh02 (approved):
Looks good, I forgot about this part!

PR #927, opened 2024-09-18
Fix charliecloud build

pagrubel:
The --force flag can cause problems, do we need to delete it or fix it in some manner? 
Deleting it fixed it on roci.


pagrubel:
This is outdated so closing

Issue #928, opened 2024-09-19
Add networkx to pyproject.toml


pagrubel:
We need to make sure that when we publish a release and it is updated in an enviroment that networkx doesn't get removed.
Add it to the pypoject.toml should fix this.

pagrubel:
Do a poetry install and push both pyproject.toml and poetry.lock for the fix.

pagrubel:
addressed in #929

PR #929, opened 2024-09-19
Issue 928 fix networkx issue

Leahh02:
I added networkx version 3.1 to the pyproject.toml. It was always working for me so it would be helpful for someone else to test it.


pagrubel (approved):
Thanks I think this will work since I did a poetry install the uninstalled networkx (got the same error when using the pip version) then did a poetry install and the error is gone. The only way to know will to merge the PR and put out a pre-release and check a pip install. Thanks!

pagrubel:
Addresses #928

PR #930, opened 2024-09-20
DEPENDS_ON arrows and text bold

Leahh02:
Made the depends on arrows and text bold to address part of Issue 921. I also realized that when I set the output label to "value" the cat-grep-tar dag was messed up, so I set the label back to "glob". I ran the generate_graph.py file on my personal laptop, so I also want to try running it on Darwin to make sure it's still fine.


pagrubel (commented):
Leah, can you make the line a little thicker?

pagrubel (commented):
@Leahh02 In addition to making the lines a little thicker for DEPENDS_ON, could you add RESTARTED_FROM.

Here is an example of a checkpointed workflow.
fd-complete

pagrubel (commented):
@Leahh02 Just noticed this output from ``beeflow dag --help``:
```

   Usage: beeflow dag [OPTIONS] WF_ID

  Export a DAG of the workflow to a GraphML file.
```
should it be png image file

pagrubel (approved):
Thanks Leah this look great! 
The help verbage will be addressed in the other dag PR

PR #933, opened 2024-10-01
Issue883/add check for different front end

### LINKED ISSUE #883: Add Check for Different Front End
Currently, if a user starts BEE on one front end of an arbitrary HPC cluster, let's call it clusterfe1, then tries to use beeflow commands on another front end of the same cluster the commands will fail since processes on different front ends usually can't communicate with each other. 

If the user tries to run a `beeflow core` command they'll get the error:
`Cannot connect to the beeflow daemon, is it running? Check the log at ".beeflow/logs/beeflow.log".`

If the user tries to use any `beeflow` commands they'll get a message like 
`Submit: Could not reach WF Manager.`

We should add a check in core.py and client.py for to make sure the host that the user is running on is the same as the host beeflow is currently running on. 
 
One big issue is there isn't currently a clean way to get this information. 

We have several options:
1. The `beeflow` log at `.beeflow/logs/beeflow.log` says the front end on which beeflow was last started in the format. 
```
Running on cluster-fe1
Launching components in order: ['redis', 'scheduler', 'celery', 'slurmrestd', 'wf_manager', 'task_manager']
```
We could grep the last Running message out of the log (and verify there wasn't a Kill operation afterwards) to get this info. This could break if we ever make changes to the beeflow log and is kind of brittle. 

2. Alternatively, we could add the hostname where beeflow is running to the workflow DB and get that information in the beeflow client. Currently, we're only using the workflow DB in the wf_manager so this would add another piece of code that depends on it which breaks our modularity somewhat. Another _issue_ is that this won't work if in the future we enable a client to run on a separate system from the one where the workflow manager is running, but that situation wouldn't be impacted by this problem so we'd need to just not do this check if we're connecting to the workflow manager from another machine. 

I think option 2 is the best solution at the moment.

kchilleri:
This PR adds a database to the client to keep track of the name of the front end where beeflow is started. Every time a ```beeflow core *``` command is submitted, we check if the current front end is the same as the one where beeflow was started from. If not, it returns a warning and exits.


kchilleri:
To Do:
1. Add more error checking.
2. Add testing (I have a test for the database but need a separate one for the rest of the code).

beeflow/client/bee_client.py
```diff
@@ -53,6 +55,37 @@ def __init__(self, *args):
...
+    """Set up front end name when beeflow core start is returned."""
+    db = bdb.connect_db(client_db, db_path())
+    db.info.set_hostname(start_hn)
+
+
+def check_hostname(curr_hn, stop=False):
+    """Check current front end name matches the one beeflow was started on."""
+    db = bdb.connect_db(client_db, db_path())
+    start_hn = db.info.get_hostname()
+    if start_hn != "":
```
rstyd: I think this would be a bit cleaner if the two ifs were combined like `start_hn != "" and curr_hn != start_hn.

beeflow/client/bee_client.py
```diff
@@ -53,6 +55,36 @@ def __init__(self, *args):
...
+    db = bdb.connect_db(client_db, db_path())
+    db.info.set_hostname(start_hn)
+
+
+def check_hostname(curr_hn, stop=False):
+    """Check current front end name matches the one beeflow was started on."""
+    db = bdb.connect_db(client_db, db_path())
+    start_hn = db.info.get_hostname()
+    if start_hn != "" and curr_hn != start_hn:  # noqa: don't use set instead
+        warn(f'beeflow was started on "{start_hn}" and you are trying to '
```
rstyd, line 87: So currently the output is: 

```
Cannot connect to the beeflow daemon, is it running? Check the log at "/home/<username>/.beeflow/logs/beeflow.log".
beeflow was started on "hostname-fe3" and you are trying to run a command on "hostname-fe4".
```
This is great, but I think we should remove the old `Cannot connect to beeflow daemon` message since we know approximately whether beeflow is up or not with this.

beeflow/common/db/client_db.py
```diff
@@ -0,0 +1,64 @@
...
+        bdb.run(self.db_file, stmt, [new_hostname])
+
+    def get_hostname(self):
+        """Return hostname for current front end."""
+        stmt = "SELECT hostname FROM info"
+        result = bdb.getone(self.db_file, stmt)[0]
+        hostname = result
+        return hostname
+
+    def get_info(self):
```
rstyd: It doesn't look like this function is getting used

beeflow/client/bee_client.py
kchilleri: Got it!

pagrubel (commented):
I was able to do a "beeflow core start" on a  front end different than one that had beeflow running on it. I don't see a check for this in beeflow start command in core.py

pagrubel (commented):
It looks like there is only a check for the commands ``beeflow core``  ``status`` and ``stop``
There should also be checks for ``start`` , ``restart``, and ``reset``.

It isn't needed for ``pull-deps`` or ``info`` though

kchilleri:
@pagrubel I am not adding checks to ```restart``` and ```reset``` commands because they call the ```stop``` function which already contains the check, but I will add one to the ```start``` command

pagrubel:
> @pagrubel I am not adding checks to `restart` and `reset` commands because they call the `stop` function which already contains the check, but I will add one to the `start` command

That makes sense for ``restart`` but ``reset`` does quite a bit of checking before it does the stop so I think a check there might be warranted

pagrubel (approved):
This works for reset and start now.

Issue #894, opened 2024-07-30
Cancelled Workflows - Archive


pagrubel:
We should probably archive Cancelled workflows and make the state Archived/Cancelled.

PR #939, opened 2024-10-03
Archive Cancelled Workflows

Leahh02:
I was debating whether I should make this call to archive the workflow in wf_actions or wf_update. The calls to archive_workflow for other states are in wf_update, but I put it wf_actions to keep the full lifecycle of actions related to cancelling the workflow in one place.

### DIFF
--- a/beeflow/client/bee_client.py
+++ b/beeflow/client/bee_client.py
@@ -643,6 +643,7 @@ def dag(wf_id: str = typer.Argument(..., callback=match_short_id),
     # Check if the workflow is archived
     wf_status = get_wf_status(wf_id)
     if wf_status == 'Archived':
+        copy_dag_in_archive = False
         bee_workdir = wf_utils.get_bee_workdir()
         mount_dir = os.path.join(bee_workdir, 'gdb_mount')
         graphmls_dir = mount_dir + '/graphmls'
@@ -650,10 +651,11 @@ def dag(wf_id: str = typer.Argument(..., callback=match_short_id),
                     "in the archive directory.",
                     fg=typer.colors.MAGENTA)
     else:
+        copy_dag_in_archive = True
         wf_dir = wf_utils.get_workflow_dir(wf_id)
         graphmls_dir = wf_dir + '/graphmls'
         os.makedirs(graphmls_dir, exist_ok=True)
-    wf_utils.export_dag(wf_id, output_dir, graphmls_dir, no_dag_dir)
+    wf_utils.export_dag(wf_id, output_dir, graphmls_dir, no_dag_dir, copy_dag_in_archive)
     typer.secho(f"DAG for workflow {_short_id(wf_id)} has been exported successfully.",
                 fg=typer.colors.GREEN)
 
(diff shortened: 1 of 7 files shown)


pagrubel (commented):
In general we have been marking states as Archived/{final_state} as they are archive. So I think we should stick to this format. I'll look to  see how this is used. I think it is also used when there is a reset.

pagrubel (changes requested):
@Leahh02 This actually ended up with the state we need for the workflow.
```
clamr	be196f	Archived/Cancelled
```
when I ran it because when a workflow is archived it automatically adds the final state.
I'm not sure what happened to the state you add here.

Also, I'd like to see any dags that were run in the graphml files as well as the final dag in the archive. 

Sorry that this wasn't asked for in the issue. I ran some dags before cancelling the workflow. There are none in the archive. I'm thinking someone may have a complex workflow that they cancelled due to something like an upcoming DST and they want to quickly see what parts actually ran.

Leahh02:
> @Leahh02 This actually ended up with the state we need for the workflow.
> 
> ```
> clamr	be196f	Archived/Cancelled
> ```
> 
> when I ran it because when a workflow is archived it automatically adds the final state. I'm not sure what happened to the state you add here.
> 
> Also, I'd like to see any dags that were run in the graphml files as well as the final dag in the archive.
> 
> Sorry that this wasn't asked for in the issue. I ran some dags before cancelling the workflow. There are none in the archive. I'm thinking someone may have a complex workflow that they cancelled due to something like an upcoming DST and they want to quickly see what parts actually ran.

I think I made the state "Cancelled and Archived" because I thought that would be reflected in the GDB. It's not, so I changed it back to the way it was for simplicity.

Leahh02:
Right now the unit tests aren't testing archiving the workflow after it's been archived. I added the line `mocker.patch('beeflow.wf_manager.resources.wf_actions.archive_workflow', return_value=None)` to test_wf_manager to skip this. I was wondering if archiving should be in a different test.

Leahh02:
all of the generated DAGs are now also being saved to the archive

pagrubel:
Now the workflow is not actually being cancelled, instead it actually runs to completion:
beeflow query 0a
Archived/Cancelled
clamr--RUNNING
ffmpeg--WAITING

$ beeflow query 0a
Archived/Cancelled
clamr--COMPLETED
ffmpeg--PENDING

$ beeflow query 0a
Archived
clamr--COMPLETED
ffmpeg--COMPLETED

I didn't have time to look, but most likely the wf_manager is looking for the Cancelled state to stop it from running.

pagrubel:
@Leahh02 This is not caused by your PR because it also occurs in the devlop branch.

Leahh02:
> @Leahh02 This is not caused by your PR because it also occurs in the devlop branch.

That's so strange. I noticed it once while I was running a workflow, but I canceled it while tasks were completing, so I figured that's why it finished. That's definitely not the case in the example you gave.

I'll try to figure it out

pagrubel (approved):
This looks good. Cancelling is not working but is another issue #955 to be addressed. This addresses #894

PR #940, opened 2024-10-03
Remove gdb ports from config

Leahh02:
Removed the ports from the gdb config section. I also removed them from lines 219-222 because they weren't being used anywhere else, but I can add those lines back in if I shouldn't have done that.


pagrubel (approved):
This looks good, thank you Leah!

PR #941, opened 2024-10-07
Do not allow beeflow to run on a compute node

### LINKED ISSUE #932: Do not allow ```beeflow core start``` from compute node
If environment variables SLURMD_* exist, then you are on a compute node. Do not allow beeflow to start and print out warning.

kchilleri:
We check the environment variables at start. If we find the variable ```SLURM_JOB_NODELIST``` then it means a node has been allocated with the Slurm scheduler. We print a message, exit out, and not allow beeflow to start.


pagrubel (approved):
Thanks @kchilleri this looks good and works well.

pagrubel:
Satifies #932

PR #945, opened 2024-10-10
Issue943/pull deps neo4j

Leahh02:
- [x] beeflow core pull-deps should use a Dockerfile to build the needed neo4j container and add the apoc file
- [x] This dockerfile should be placed in beeflow/data/dockerfiles and should be used by the ci integration tests so we only need to modify it in one place.
- [x] Also, if a user uses "beeflow core pull-deps" to have their own dependent containers, they should be used in the bee.conf file. Consider automatically editting it and removing the ~/.beeflow/deps directory. Beeflow should not be running when this is done, so adding a check for that should probably be done.


Leahh02:
I'm working on the last part right now. I can change the name of beeflow/data/dockerfiles/Dockerfile.apoc_neo4j if needed. I can also remove beeflow/data/dockerfiles/Dockerfile.neo4j-ppc64le if needed.

pagrubel:
> I'm working on the last part right now. I can change the name of beeflow/data/dockerfiles/Dockerfile.apoc_neo4j if needed. I can also remove beeflow/data/dockerfiles/Dockerfile.neo4j-ppc64le if needed.

Yes, I think Dockerfile.neo4j would suffice and yes please remove the other. We will need to probably add one for arm architecture in the future.

beeflow/client/core.py
```diff
@@ -35,6 +37,9 @@
 from beeflow.common.deps import redis_manager
 
 
+REPO_PATH = Path(*Path(__file__).parts[:-3])
```
pagrubel, line 40: Will this work for beeflow when published in PyPi?

pagrubel (approved):
This is a good refactor of utilities and tests used while incorporating pulling dependent containers used for BEE. It is a nice feature to update the config file to use the newly pulled dependent containers. My only concern is how the path to the Dockerfile will work once published in pip. In order to test that I'm going to merge this and publish it in a pre-release.

rstyd:
Looks good to me!

pagrubel:
Just a note. If someone has already started beeflow with other dependent containers, it will continue to use them until they delete the bee_workdir/deps or perform a beeflow core reset, but this can be another issue.

PR #947, opened 2024-10-11
Allow beeflow to run on back end node

### LINKED ISSUE #946: Add flag to allow beeflow to run on compute node
Currently, we do not allow beeflow to run on Slurm allocated nodes. This interrupts CI testing. We need to add a flag, say ```beeflow core start --compute-node-allow``` or ```beeflow core start --backend```, that allows beeflow to run on backend nodes for testing purposes.

kchilleri:
Adding a flag to ```beeflow core start``` to allow ```beeflow``` to run on a back end node for testing purposes.

We can now run with the following flag: ```-B, --backend [default: False]```.

Documentation has also been updated.

### DIFF
--- a/beeflow/client/core.py
+++ b/beeflow/client/core.py
@@ -303,7 +303,6 @@ def check_dependencies(backend=False):
         warn('Slurm job node detected! Beeflow should not be run on a compute node.')
         warn(f'SLURM_JOB_NODELIST = {os.environ.get("SLURM_JOB_NODELIST")}')
         sys.exit(1)
-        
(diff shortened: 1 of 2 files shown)


pagrubel (approved):
The change looks good to me. I tried it out on darwin and it still warns me if on a backend node. When I tried with the flag it did attempt to start beeflow, however many of the components failed.

pagrubel:
This was a problem with my environment, all looks good. The flag is really just for testing purposes and should not be used by users.

Issue #772, opened 2024-01-30
Change Password Mechanism


rstyd:
We need to change the way passwords are handled in Neo4J (if we choose to stick with it). Currently passwords are set in the configuration file and used in start_gdb. I propose we modify this to create an authorization code file when beeflow first starts then the configuration file will have the path to this file.

aquan9:
I made PR #948 to fix this issue by generating a random base64 value for the password that's stored in the user's home directory config file on generation of the config file.

pagrubel:
We had to revert the changes from this PR since it caused the issue reported in #959 so I'm re-opening this issue

rstyd:
Need to additionally make sure we no longer hardcode the neo4j password in the integration test.

rstyd:
Need to rethink this. Is a password really the best option?

PR #948, opened 2024-10-15
Use random pass for neo4j

aquan9:
This PR changes the way that we initialize the default password for the neo4j database to be based on a random base64 hash value rather than just the default "password" text, which should hopefully make our database more secure on shared systems. 

The concern is that since the graph database is setup on the shared front-end, automatically setting up a default password of "password" would let other users explore things like the names of steps in another user's workflow, and how their workflow is setup.


aquan9:
This PR resolves issue #772

pagrubel (approved):
This looks good, thanks @aquan9. I merged develop and fixed conflicts caused by a different change to configuration.

Issue #949, opened 2024-10-16
beeflow config new doesn't work if there's no config file


rstyd:
We currently have an issue where if someone runs `beeflow config new` without an existing config file, beeflow exits with `Configuration file does not exist! Please try running beeflow config new`.  This completely breaks BEE for new users since they can never generate a config file.

Issue #935, opened 2024-10-01
Infer Slurmrestd Version


rstyd:
We currently have to deal with shifting slurmrestd versions whenever slurm gets updated. We should infer the newest slurmrestd version using `slurmrestd list -s`.

PR #950, opened 2024-10-16
Fix config issue and revise slurmrestd inference

rstyd:
This PR addresses #949 and #935.


aquan9 (approved):
This looks good to me

PR #953, opened 2024-11-04
Fix slurmrestd version function

rstyd:
This PR fixes the slurmrestd inference function. It also adds a debug message that states what version it finds.


pagrubel (commented):
@rstyd Should there be a fix for the configuration file?

pagrubel:
Please remove api version from config and use whatever is installed.

pagrubel:
I tried this out with and without slurmrestd and it seems to work well. I haven't reviewed the changes yet.

kchilleri (approved):
This also works for me + I can find the openapi_version in the beeflow logs.

Issue #951, opened 2024-10-29
Remove excess output from `beeflow config new`


Leahh02:
Right now if you run `beeflow config new` and you don't have an existing config file you'll get a bunch of output after the config file is saved.
Here's an example that's edited down:
```
[2024-10-29 10:39:16,045] beeflow.common.config_validator:_check_required_options(): option DEFAULT::bee_workdir is missing, setting to default "/path/to/.beeflow"; please check that that is correct and update the config
Traceback
beeflow.common.config_validator.ConfigError: Config validation failed:
    * option DEFAULT::workload_scheduler is required but has not been set
    * option DEFAULT::neo4j_image is required but has not been set
    * option DEFAULT::redis_image is required but has not been set
    * option job::default_account is required but has not been set
    * option job::default_time_limit is required but has not been set
    * option job::default_partition is required but has not been set
```

I commented out lines 19-27 of beeflow/common/deps/neo4j_manager.py and lines 9-11 of beeflow/common/gdb/graphml_key_updater.py and I wasn't getting the output anymore.  Maybe these lines should be moved inside a function so they're not run when the modules are imported.

Leahh02:
I think lines 100-105 of beeflow/common/config_driver.py also need to be put in a "while true" block, so it would look like:
```
        while True:
            try:
                with open(USERCONFIG_FILE, encoding='utf-8') as fp:
                    config.read_file(fp)
                break
            except FileNotFoundError:
                print("Configuration file is missing! Generating new config file.")
                new(USERCONFIG_FILE)
```

pagrubel:
Closing this as it has been fixed

PR #954, opened 2024-11-05
fix excess `beeflow config new` output

Leahh02:
Addresses issue #951. Let me know if the way I did the global variables in beeflow/common/deps/neo4j_manager.py is a bad practice, I just wanted to keep most of the code the same.


rstyd (approved):
Looks good to me!

PR #957, opened 2024-11-08
Check for default dependency containers for new configuration

pagrubel:
Checks if default dependency containers exist and will automatically put their paths in bee.conf if they exist.

If a different container is desired the user will have to edit the bee.conf file manually.


pagrubel:
Addresses #957

rstyd (approved):
It works. Thanks!

Issue #955, opened 2024-11-06
Cancel Workflows not working


pagrubel:
Cancelling workflow is not working. At one point we let running or schedule tasks complete and just did not schedule any new ones but now all tasks continue. When testing this, I ran beeflow-dag several times and it only archived the final one, so this will also need to be tested when cancel is fixed.

The workflow is not actually being cancelled, instead it actually runs to completion:
beeflow query 0a
Archived/Cancelled
clamr--RUNNING
ffmpeg--WAITING

$ beeflow query 0a
Archived/Cancelled
clamr--COMPLETED
ffmpeg--PENDING

$ beeflow query 0a
Archived
clamr--COMPLETED
ffmpeg--COMPLETED

Leahh02:
I realized that whenever I cancel a workflow I always pause it beforehand, so I think maybe that's why I've only seen his be a problem once.

I looked into it, `beeflow cancel` uses `conn.delete(_resource(long_wf_id), json={'option': 'cancel'}, timeout=60)`. In beeflow/wf_manager/resources/wf_actions.py the docstring for  the delete method says that "For cancel, current tasks finish running." but that shouldn't mean that tasks that haven't started yet should start.

pagrubel:
So, currently scheduled jobs should be allowed to complete, but no others should be scheduled. We have discussed an option to cancel all jobs too, but for sure tasks that are waiting should not run.

Leahh02:
> So, currently scheduled jobs should be allowed to complete, but no others should be scheduled. We have discussed an option to cancel all jobs too, but for sure tasks that are waiting should not run.

That makes sense

pagrubel:
Oh and the Archived/Cancelled state should not change to Archived, and of course all dags that were done should be in the archive.

pagrubel:
Addressed by #960

PR #960, opened 2024-11-21
Fix Cancel Workflows

Leahh02:
This PR fixes cancel workflows. When a workflow was cancelled it ran to completion. Now running or scheduled tasks complete, but new tasks aren't scheduled.


Leahh02:
I don't have the testing mentioned in Issue #955. I was thinking that should maybe be in a separate PR so this one isn't overwhelming.

pagrubel:
> I don't have the testing mentioned in Issue #955. I was thinking that should maybe be in a separate PR so this one isn't overwhelming.

That's fine. Let's make another issure for that so we can close #955 when this is merged

pagrubel (approved):
Good work!
For now if a workflow is cancelled we don't cancel jobs. I like how if all the jobs are already running the workflow continues and is archived once they are completed. And any running jobs are archived. We will need to look at testing and think about how we might want to allow users the option to cancel jobs.

PR #965, opened 2024-12-04
Issue877/Add config option for archive directory

### LINKED ISSUE #877: Add configuration option to specify the archive directory
@kabir-vats suggested this. _Should_ be pretty straightforward.

arhall0:
Closes #877 

This PR adds a config option to change the directory where workflows are archived. By default this is set to the same location that was used previously (`~/.beeflow/archives`). If you manually adjust the `bee.conf` option `bee_archive_dir` to any other folder; it will be used instead. As with the default, the folder does not need to already exist.

I tested this using the `cat-grep-tar` example ensuring the default behavior worked as expected. I then set `bee_archive_dir = $HOMEDIR/BEE_jobs/archives` with $HOMEDIR filled in as appropriate; and ensured the archived was placed in the new folder.

I ensured tests passed using:

```shell
pytest beeflow/tests
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -56,6 +56,7 @@ Contributors:
 * Patricia Grubel - `pagrubel <https://github.com/pagrubel>`_
 * Qiang Guan - `guanxyz <https://github.com/guanxyz>`_
 * Ragini Gupta - `raginigupta6 <https://github.com/raginigupta6>`_
+* Aaron R Hall - `arhall0 <https://github.com/arhall0>`_
 * Leah Howell - `Leahh02 <https://github.com/Leahh02>`_
 * Andres Quan - `aquan9 <https://github.com/aquan9>`_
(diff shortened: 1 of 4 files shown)


pagrubel (commented):
Can you add some documentation on how to use this? Should it be a configuration option?

rstyd:
Now that you mention it, it doesn't look like we currently have anything in the documentation that breaks down the individual configuration options, and I can't find a good place to include this info in the documentation as currently written.  I propose we create a separate issue to add a page to the docs that breaks down all the config options and it'll include this as well.

pagrubel:
Sounds good

rstyd:
This looks great! The only change I would recommend is to add a line to ci/bee_config.sh so that it's using your new config option. I'm also putting up an issue for changing this script so it's not hard coded since it's a pain to maintain.

rstyd (approved):
LGTM! Approved!

PR #973, opened 2024-12-09
Correct archives location

arhall0:
Follow-up related to issue #877. 

I checked `git grep archives` on `develop` and noticed some mention of the archives folder that should be updated to use the location defined in the config:

```shell
beeflow/client/core.py:    archive_dirs = ['logs', 'container_archive', 'archives', 'workflows']
beeflow/wf_manager/resources/wf_actions.py:            archive_path = f"{bee_workdir}/archives/{wf_id}.tgz"
beeflow/wf_manager/resources/wf_list.py:        archive_path = os.path.join(bee_workdir, 'archives', wf_id + '.tgz')
```

### DIFF
--- a/beeflow/client/core.py
+++ b/beeflow/client/core.py
@@ -559,6 +559,7 @@ def reset(archive: bool = typer.Option(False, '--archive', '-a',
         Shutdown beeflow and all BEE components.
         Delete the bee_workdir directory which results in:
             Removing the archive of all workflows.
+                (unless archives is configured elsewhere).
             Removing the archive of workflow containers
                 (unless container_archive is configured elsewhere).
             Reset all databases associated with the beeflow app.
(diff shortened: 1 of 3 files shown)


arhall0:
The mention in `beeflow/client/core.py` is concerning making an archive of some folders in `.beeflow` when using `beeflow core reset --archive`. After discussion with @rstyd we decided it didn't make sense to add the workflow `archives` folder to this `.beeflow` archive if it was configured to be elsewhere. A message was added about this when running `beeflow core reset` inline with changes to `container_archive`.

The mention in `beeflow/wf_manager/resources/wf_actions.py` concerns the command `beeflow remove` which is for removing a specific workflow. This was adjusted so the archive of that workflow will be deleted as appropriate regardless of its location.

The mention in `beeflow/wf_manager/resources/wf_list.py` concerns the command `beeflow copy`. Although I adjusted this as appropriate to use the `archives` location as specified in the config file; this command currently appears bugged and may be removed per discussion with @rstyd.

rstyd (approved):
This looks great! Approved.

PR #980, opened 2024-12-16
Add option to archive the workdir

### LINKED ISSUE #900: Workdir Should be Added to the Archive (Potentially)

arhall0:
Adds the option to archive the `workdir` when using `beeflow submit` by adding the `--archive-workdir` flag. Checks the `workdir` isn't too large: too large file size (>1000 MB), contains too many files (>10,000), contains too many directories (>1000). If it is, prompts user to verify.

Could potentially add a config option for this to always attempt to archive the `workdir`. Currently that isn't implemented.

### DIFF
--- a/beeflow/client/bee_client.py
+++ b/beeflow/client/bee_client.py
@@ -242,7 +242,9 @@ def submit(wf_name: str = typer.Argument(..., help='the workflow name'),  # pyli
            workdir: pathlib.Path = typer.Argument(...,
            help='working directory for workflow containing input + output files',),
            no_start: bool = typer.Option(False, '--no-start', '-n',
-                                         help='do not start the workflow')):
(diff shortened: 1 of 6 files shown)


beeflow/client/bee_client.py
```diff
@@ -312,6 +314,51 @@ def is_parent(parent, path):
     if os.path.commonpath([os.path.realpath('/var/tmp'), workdir]) == os.path.realpath('/var/tmp'):
         error_exit("Workflow working directory cannot be in \"/var/tmp\"")
 
+    # TODO: this should be in a function somewhere
```
arhall0, line 317: I tried to find some kind of python package that would do this logic, but didn't come up with anything.

I'm not sure the best place to put this function. It should likely also have some tests.

PR #981, opened 2024-12-16
Fix backend issues and reset client db

### LINKED ISSUE #974: There is a problem if beeflow is started on a compute node and that allocation is lost or an error occurs and it is exited before beeflow is stopped
Our overnight test were having a problem with this. Maybe there is a way to check if the comput node is actually allocated to the user. So when our database shows beeflow is running somewhere that we don't even have access to it causes an error.

kchilleri:
This PR fixes issue #974.

**client_db** stores the following variables:
   _hostname_ of the front end/back end beeflow was started from, default is empty
   _backend_ flag: default is empty, "true" if beeflow was started on a compute node

When doing ```beeflow core start```: client database is populated (--backend, when appropriate).
When doing ```beeflow core stop```: client database gets reset, back to default values.

The issues appeared when ```beeflow core stop``` was not used and therefore the client database was not reset. I have added some checks for all cases I could think of when this happens. For example, If beeflow is started on a compute node and the allocation ends, or is relinquished, I have added checks every time ```beeflow core start``` is returned that check the database values and resets them accordingly.

Also, I added an error check that checks if the user is using the ```--backend``` flag when not on a backend node.

A test for the client database has been updated as well.

### DIFF
--- a/beeflow/client/core.py
+++ b/beeflow/client/core.py
@@ -407,7 +407,7 @@ def start(foreground: bool = typer.Option(False, '--foreground', '-F',
           '-B', help='allow to run on a backend node')):
     """Start all BEE components."""
     start_hn = socket.gethostname()  # hostname when beeflow starts
-    if bee_client.get_hostname() == "" and bee_client.check_backend_status() == "":  # noqa
+    if bee_client.get_hostname() == "" and bee_client.check_backend_status() == "":
         bee_client.setup_hostname(start_hn)  # add to client db
     else:
         bee_client.check_db_flags(start_hn)
--- a/beeflow/client/core.py
+++ b/beeflow/client/core.py
@@ -305,11 +305,22 @@ def load_check_charliecloud():
 
 def check_dependencies(backend=False):
     """Check for various dependencies in the environment."""
-    # Check if running on compute node under Slurm scheduler
+    # Note: Code block of lines 309-320 is for SLURM scheduler!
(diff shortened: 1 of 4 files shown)


pagrubel (changes requested):
I tested this by allocating a compute node and starting beeflow, making sure it was running then releasing the allocation without stopping beeflow. 
Repeating the above from another compute node works, it clears up the data base and enables beeflow to be started on the new compute node.
However, if I did the above and then tried to start from the front end it does not clear up the database, instead it just fails.

beeflow/client/core.py
```diff
@@ -410,6 +410,7 @@ def start(foreground: bool = typer.Option(False, '--foreground', '-F',
     if bee_client.get_hostname() == "" and bee_client.check_backend_status() == "":
         bee_client.setup_hostname(start_hn)  # add to client db
     else:
+        time.sleep(5)  # giving Slurm time to relinquish compute node, if necessary
```
pagrubel: Please make this 10 seconds, it won't affect most users, thanks!
That works and everything else looks good and works well!
Thanks for helping me find out why it wasn't working for me. The logic seemed fine, I just did not take into consideration slurm's timing.

pagrubel (approved):

pagrubel (approved):
Great@ thank you for this work, this solves the problem when overnight tests fail but left the compute node as running beeflow in the database.

PR #982, opened 2024-12-20
Issue979/Remove metadata command

### LINKED ISSUE #979: Remove the metadata command from the beeflow client
@jtronge Added the metadata command to the beeflow client to support his GUI client. We deprecated the GUI client earlier this year, and the problem the metadata command was intended to solve (figuring out which bolt port belongs to which workflow). Isn't an issue now since all workflows share a bolt port due to @kabir-vats refactor of the GDB. I propose we remove command and the associated resource in `beeflow/wf_manager/resources`.

arhall0:
Primarily this PR reverts commit e7778c754234cd1706da2fbb4521e2f2c6403be2 which added the metadata command to the beeflow client. There was one line change on that commit which appeared unrelated which was not reverted.

I grep'd the repo for 'metadata' to see if I could find any related code changes since the command was added, but didn't see anything of note.

`pytest beeflow/tests` passes without issue.


rstyd (approved):
Looks good to me!

PR #983, opened 2024-12-24
Issue963/Make beeflow config new non-interactive by default

### LINKED ISSUE #963: Non-interactive "beeflow config new"
We should be able to run "beeflow config new" without interactions. Would we still want the interactive mode?
What would the defaults be? Most likely Slurm and blank times, and optional values for slurm like qos, account, reservation and time-limit. How would it be run for Flux?

Maybe we can have a flag for Flux, keeping in mind that we may need to add another workload scheduler like LSF
beeflow config new --workload_scheduler="Flux"

I think we need to look into the slurm values. What can we already add per job?

Also, the current "beeflow core start" and "beeflow list" now automatically check for a bee.conf and if it doesn't exist it makes one so that will need to be thought about.

arhall0:
Closes #963 

This PR makes `beeflow config new` and automatic incantations non-interactive by default. All config options now have a default value which is used for the non-interactive mode. `beeflow config new --flux` overrides `workflow_scheduler` default of "Slurm" to be "Flux".

`beeflow config new --interactive` will prompt the user for every config option, but entering nothing will now use the stored default.

### DIFF
--- a/beeflow/common/config_driver.py
+++ b/beeflow/common/config_driver.py
@@ -241,36 +241,39 @@ VALIDATOR = ConfigValidator('BEE configuration file and validation information.'
 VALIDATOR.section('DEFAULT', info='Default bee.conf configuration section.')
 
 VALIDATOR.option('DEFAULT', 'bee_workdir', info='main BEE workdir',
-                 default=DEFAULT_BEE_WORKDIR, validator=validation.make_dir)
+                 default=DEFAULT_BEE_WORKDIR, validator=validation.make_dir,
+                 prompt=True)
 
 VALIDATOR.option('DEFAULT', 'bee_archive_dir', info='directory to store workflow archives',
-                 default=DEFAULT_BEE_ARCHIVE_DIR, validator=validation.make_dir)
+                 default=DEFAULT_BEE_ARCHIVE_DIR, validator=validation.make_dir,
+                 prompt=True)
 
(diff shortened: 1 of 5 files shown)


pagrubel (changes requested):
@arhall0 The default is great, makes it so much easier to set up. However, I don't think the interactive mode should go through all the attributes, its just too confusing and some of those values are really for developer use only. Also if someone chooses Flux in that mode it doesn't eliminate the slurm section, That's just one example. Maybe you could set up a dictionary of what attributes we would allow them to change this way similar to the old behavior.

pagrubel:
Here are a list of attributes I think should be queried in the interactive mode:

bee_workdir 
bee_archive_dir 
neo4j_image
redis_image
max_restarts

default_account 
default_time_limit 
default_partition 

container_archive 

Also, There should be a way to do an interactive session that works for flux. When I tried both flags it didn't work well. It did change to flux but still added the slurm section.

arhall0:
I also added `workflow_scheduler` to be prompted during `--interactive`. This is overridden if the `--flux` is used.

This was accomplished by adding a `prompt` argument to `config_validator.ConfigOption` that can be set for each option.

pagrubel (approved):

pagrubel:
Thank you this look great!

pagrubel (approved):

PR #990, opened 2025-01-17
beeflow:useContainer support for SquashFS

### LINKED ISSUE #967: Add capability to run applications that are Charliecloud sqfs format (DockerRequirement) for tasks
This should be like the Container use functionality although we don't need to do a convert to tmp space.
However, there may need to be a check that squash fuse is installed. This may already be tested by Charliecloud can we just pass on the error? 

Maybe we can put a test that fails in our integration tests and test for the message and then test overnight on our systems that we know have squash fuse.

arhall0:
Closes #967 

Enables the use of `SquashFS` format for the `DockerRequirement` `beeflow:useContainer`.

I tested this by modifying the clamr example's `clamr_wf.cwl` file and changed the two `DockerRequirement`s to be:

```
        DockerRequirement:
            beeflow:useContainer: "$WORKDIR_PATH/clamr-ffmpeg.sqfs"
```

Where `clamr-ffmpeg.sqfs` was made using the `Dockerfile` included with the example.

### DIFF
--- a/beeflow/common/crt/charliecloud_driver.py
+++ b/beeflow/common/crt/charliecloud_driver.py
@@ -101,9 +101,14 @@ class CharliecloudDriver(ContainerRuntimeDriver):
             container_path = '/'.join([self.container_archive, task_container_name]) + '.tar.gz'
 
         # If use_container is specified, no copying is done, the file  path is used
+        squashfs = False
         if use_container:
             task_container_name = self.get_ccname(use_container)
             container_path = os.path.expanduser(use_container)
+            _, ext = os.path.splitext(container_path)
+            # infer if using SquashFS, similar to:
+            # https://hpc.github.io/charliecloud/ch-convert.html#format-inference
+            squashfs = ext in ['.sqfs', '.squash', '.squashfs']
         else:
             container_path = '/'.join([self.container_archive, task_container_name]) + '.tar.gz'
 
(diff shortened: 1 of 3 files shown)


arhall0:
I'm currently not checking if `SquashFS` is installed. So an error would be thrown from the job by `Charliecloud` and end up in the archive. I don't have a good way to check this situation. 

We could throw an error earlier with something like `if shutil.which('fusermount3') is None` then error based on https://hpc.github.io/charliecloud/ch-run.html#image-format

beeflow/common/crt/charliecloud_driver.py
```diff
@@ -101,9 +101,14 @@ def run_text(self, task):  # noqa
             container_path = '/'.join([self.container_archive, task_container_name]) + '.tar.gz'
```
pagrubel:
```suggestion
            container_path = '/'.join([self.container_archive, task_container_name]) + '.tar.gz'
```
Is this line needed anymore?

pagrubel:
> I'm currently not checking if `SquashFS` is installed. So an error would be thrown from the job by `Charliecloud` and end up in the archive. I don't have a good way to check this situation.
> 
> We could throw an error earlier with something like `if shutil.which('fusermount3') is None` then error based on https://hpc.github.io/charliecloud/ch-run.html#image-format

Could this be tested in our github ci, I don't think we have Squash fuse installed.

arhall0:
> Could this be tested in our github ci, I don't think we have Squash fuse installed.

I made a debug test to run in CI to see if `fusermount3` is available and it seems like it is:
```shell
 =================================== FAILURES ===================================
_________________________ test_debug_check_fusermount3 _________________________

    def test_debug_check_fusermount3():
        """Debug check if system has `fusermount3`."""
>       assert shutil.which('fusermount3') is None
E       AssertionError: assert '/usr/bin/fusermount3' is None
E        +  where '/usr/bin/fusermount3' = <function which at 0x7f5cc7b41990>('fusermount3')
E        +    where <function which at 0x7f5cc7b41990> = shutil.which

beeflow/tests/test_charliecloud_driver.py:76: AssertionError
```

I believe `fusermount3` is all that `Charliecloud` needs to use `SquashFS` for `ch-run`, although I could be mistaken.

arhall0:
Note: Unit tests failure in the CI right now seems to be when its trying to update the coverage badge. This branch is a bit out of date with `develop` and that seems to break that process until you merge/rebase.

pagrubel:
I think we can let Charliecloud handle what happens if the requirement for sqfs doesn't exist. I just had an error. It did it fairly well when I tried the .sqfs conatiner using the develop branch. The clamr task failed and the workflow failed due to that. I was able to find the errors in the stderr file.

pagrubel:
This looks good, can you get the unit tests cleared up?

pagrubel (approved):
This works well and looks great. Thanks for adding the test.

PR #996, opened 2025-01-27
Change SchedulerRequirement to SlurmRequirement; add qos and reservation

### LINKED ISSUE #989: Add capabilities to beeflow:SchedulerRequirement
Currently, SchedulerRequirement includes account, time-limit and partition. We also need reservation and qos and should determine if others would be helpful.
These should be in config, with default as blank as well.

Also, should we consider changing the name of the requirement to WorkloadManagerRequirement (too long) since this is different than the BEE Scheduler component.

pagrubel:
Closes #989 
This pull request renames the beeflow:SchedulerRequirement for job requirements to the slurmworker with the better descriptive beeflow:SlurmRequirement, and adds defining qos and reservation as requirement attributes.


rstyd (approved):
This looks great! Approved

Issue #873, opened 2024-07-03
Add More Tests to improve Code Coverage


rstyd:
We need to improve BEE's code coverage. I think the best approach is to do this module by module. 
For now, let's focus on:

- [ ] Workflow Manager
- [ ] Task Manager
- [ ] Client
- [ ] Container runtime driver
- [ ] BEEConfig

pagrubel:
Make a new issue for each of the above as you work on it. They should probably be separate PR's.

Leahh02:
Andres and I thought this would be a good issue to work on to learn about code testing. I'll be starting by looking at code that I wrote that doesn't have a lot of coverage

PR #997, opened 2025-01-27
Adds regression test for archive_workflow in wf_update

arhall0:
As part of #873 I wrote a regression test for the `wf_update.archive_workflow`.


arhall0:
Made a minor adjustment to also test `archive_fail_workflow` which is a wrapper function of `archive_workflow`.

kchilleri (approved):
This tests ```archive_workflow``` and ```archive_fail_workflow``` in ```wf_update.py```. Looks good to me!

PR #1006, opened 2025-02-10
Add ScriptRequirement to Workflow module and comd example

rstyd:
This PR adds the slurm requirement to the Workflow module so users can specify a pre and post script for their workflows and adds an example pre-script to the comd workflow.


pagrubel (approved):
This looks good.

PR #1007, opened 2025-02-10
Set up unique remote port number as the default

### LINKED ISSUE #988: Replace remote_port = 7777 in config file with unique port number
Remove default number of 7777 and replace it with unique user port number within the config file.

kchilleri:
Changing default remote port number of '7777' to a unique number in the range of (7777, 16000) using the user's $UID number.

Note: This new _unique_ port number still goes through the ```is_port_taken()``` and ```find_free_port()``` functions in ```beeflow/remote.py``` in case it happens to be taken.

### DIFF
--- a/beeflow/common/config_driver.py
+++ b/beeflow/common/config_driver.py
@@ -203,6 +203,13 @@ def filepath_completion_input(*pargs, **kwargs):
         return input(*pargs, **kwargs)
 
 
+def unique_port():
+    """Assign unique port for remote user."""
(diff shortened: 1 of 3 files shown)


pagrubel (approved):
This looks good to me and gives a unique port number when issuing `beeflow config new`

rstyd (approved):
This looks great approved! We might still want a backup plan later if that port is used when the remote API gets setup but that's another issue.

Issue #969, opened 2024-12-06
Add job running limit in config for the task manager


pagrubel:
We add a configuration such as active_job_limit in the configuration for task manager. The task manager should make sure that only that number of jobs are active on the system (Pending, Running or Completing).
The task manager only allows that many jobs on the run queue, so the rest stay in the submit queue until there is space. 

We may need to add a task state for this.

pagrubel:
After discussing this, we will add a limit for jobs that are being processed at one time. The limit will be in the user's config with a default of 30.

pagrubel:
Addressed in #1010

PR #1010, opened 2025-02-12
Add job_limits to config;  limit the number jobs in the job queue

pagrubel:
Addresses #969 

This implementation gives the task the state 'SUBMIT', however when queried it remains 'WAITING'. 
So I need to figure out how to propagate the state to the query.
Also, this is a simple way to limit jobs.
In the future we might want to consider monitoring accounts. 
If a workflow is paused anything in the submit queue will continue to be scheduled. This needs to be addressed possibly in another PR.


pagrubel:
I should add the 'SUBMIT' indicates the task is in the submit queue waiting to be scheduled and placed in the job queueu.

pagrubel:
Oh just thought of this, the default is 2 right now. I need to change that to blank and make it a large number.

Leahh02 (approved):
This looks good to me!

It satisfies Issue #969. The task manager limits the number of jobs in the job queue to the number specified in the config. Tasks in the submit queue stay there until the job queue has space.

The SUBMIT state shows up in the task manager log. I think the state showing up when the workflow is queried should probably go in a separate PR so this work can be merged in.

I think it would be a good idea to have a simple unit test for this. That can also go in a separate PR.

pagrubel:
Also remaining pylint problems are being addressed in another PR

PR #1013, opened 2025-02-20
Add `__init__.py` files for `pytest-cov` discovery

arhall0:
Noticed when looking at #977 that some files are not being discovered by `pytest-cov` giving an artificially high line coverage. Adding `__init__.py` files to all the module folders forces `pytest-cov` to check files for coverage.

If there is some reason not to add these or if I missed locations that `pytest-cov` should check let me know.

### DIFF
--- a/beeflow/__init__.py
+++ b/beeflow/__init__.py
@@ -1,0 +1,1 @@
+"""beeflow module."""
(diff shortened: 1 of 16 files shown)


pagrubel (approved):
Thank you for fixing this. We need to have the additional coverage measurement so we can be aware of what is not being tested and improve our testing.

PR #1015, opened 2025-02-25
Add --remote flag to beeflow core start

### LINKED ISSUE #992: Remove remote variables from config file and add --remote flag
This will help the user when starting beeflow remotely to not have to worry if the config file is set up correctly for a remote start.
This flag will automatically assign a unique port number to use remotely.

kchilleri:
This will add a flag, ```--remote``` or ```-R```, to ```beeflow core start``` to allow for remote interaction.
All this does is replace the ```remote_api``` variable from the config file. If the ```--remote``` flag is not specified when starting ```beeflow```, then the REST API will not be initiated. This will be convenient for a user trying to interact with a ```beeflow``` instance (remotely) without having to worry if the config file is set up for a remote connection.

This closes issue #992.

### DIFF
--- a/beeflow/client/core.py
+++ b/beeflow/client/core.py
@@ -51,11 +51,7 @@ class ComponentManager:
         """Return a decorator function to be called."""
 
         def wrap(fn):
-            """Check to see if any components are disabled."""
-            if bc.get('DEFAULT', 'remote_api') is False and 'remote_api' in name:
-                return
-
-            # Add the component to the list.
+            """Add the component to the list."""
             self.components[name] = {
                 'fn': fn,
                 'deps': deps,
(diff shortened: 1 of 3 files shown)


kchilleri:
To Do: add documentation about this new flag

arhall0 (approved):
- Deleted and remade config on this branch.
- Started `beeflow` with `beeflow core start`, then checked `curl $SSH_TARGET:$API_PORT/` failed as expected.
- Tried `beeflow core start --remote`, and `curl $SSH_TARGET:$API_PORT/` was successful.

The diff looks clean and tests passed without issue. It would be nice if we could write a regression test that does something like above; but I'm not sure an easy way to do that.

pagrubel:
> * Deleted and remade config on this branch.
> * Started `beeflow` with `beeflow core start`, then checked `curl $SSH_TARGET:$API_PORT/` failed as expected.
> * Tried `beeflow core start --remote`, and `curl $SSH_TARGET:$API_PORT/` was successful.
> 
> The diff looks clean and tests passed without issue. It would be nice if we could write a regression test that does something like above; but I'm not sure an easy way to do that.

@arhall0 I'm adding this as an issue

PR #1021, opened 2025-03-06
Start on pylint fixes uncovered recently

pagrubel:


arhall0 (approved):
I ran cat-grep-tar and did `beeflow dag` a few times while it ran and they all appeared in the archive. Workflow finished without issue. I ran the ci tests and saw no issues. I also ran `ci/integration_test.py` and all passed.

The diff looks clean. Looks good!

PR #1022, opened 2025-03-06
Adjust workflow end states to include Archived/Partial-Fail

### LINKED ISSUE #1018: Adjust workflow end states
End workflow state should be `Archived`, `Archived/Failed`, or `Archived/Partial-Fail` depending on state of final task nodes.

This would prompt the user appropriately to further investigate the logs of a workflow.

arhall0:
Closes #1018 

PR #1008 removed `Archived/Failed` workflow end state. This adds back that state as well as implements `Archived/Partial-Fail` state.

The final state of a workflow is determined by checking the state of the final task nodes. If they are all successful the final state will end up as `Archived`. If they are all failed, the state will be `Archived/Failed`. If only some are failed, it will be `Archived/Partial-Fail`.

This adds an integration test for the `Archived/Partial-Fail` case.

It would be good to add some regression tests before merging.


pagrubel (approved):
Good work now the user will be able to git more information when part of a workflow fails and be flexible in their designs that they know may fail.

PR #1036, opened 2025-04-04
Fail dependent tasks for TIMEOUT

### LINKED ISSUE #1034: TIMEOUT getting stuck
When a task is in the TIMEOUT state and they don't have a task_checkpoint the task will just sit in the TIMEOUT state for forever and the dependent tasks will remain in the WAITING state. I'm assuming this is also the case for the TIMELIMIT state. 

What happens is that in update_jobs in background.py, the new_job_state will eventually be TIMEOUT, if there's no task_checkpoint the TIMEOUT state will just be pushed to the tm.db update queue, wf_update will update the neo4j db and wfm.db with this TIMEOUT sate, the handle_state_change function in wf_update.py won't do anything because the state isn't 'COMPLETED' or in ['FAILED', 'SUBMIT_FAIL', 'BUILD_FAIL'] and the workflow isn't completed or cancelled. And then when update_jobs runs again it removes the task from the job_queue. Then nothing else will happen. The task will sit in TIMEOUT and the dependent tasks will sit in WAITING.

Maybe the dependent tasks should be set DEP_FAIL?

To reproduce you can set the default_time_limit in bee.conf to something short. I did 30 seconds. And then run the clamr workflow.

arhall0:
Closes #1034 

This corrects the issue for the case described in #1034:

>To reproduce you can set the default_time_limit in bee.conf to something short. I did 30 seconds. And then run the clamr workflow.

From my read of the code, it seems like this should not affect restarts; but I still need to verify this. ~~I should also check a case for `TIMELIMIT`.~~

### DIFF
--- a/beeflow/common/gdb/neo4j_cypher.py
+++ b/beeflow/common/gdb/neo4j_cypher.py
@@ -4,7 +4,8 @@ from re import fullmatch
 from beeflow.common import log as bee_logging
 
 log = bee_logging.setup(__name__)
-
+failed_task_states = ['FAILED', 'SUBMIT_FAIL', 'BUILD_FAIL', 'DEP_FAIL', 'TIMEOUT', 'TIMELIMIT']
+final_task_states = ['COMPLETED'] + failed_task_states
 
 def create_bee_node(tx):
     """Create a BEE node in the Neo4j database.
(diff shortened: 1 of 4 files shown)


arhall0:
The integration test `checkpoint_restart` works as expected. I ran with the following:

```shell
./ci/integration_test.py --tests checkpoint_restart --timeout 9999
```

Increasing the timeout was necessary as it is quite slow. This workflow has `TIMEOUT`s. Final states:

```
Archived
clamr--TIMEOUT
ffmpeg--COMPLETED
clamr(1)--TIMEOUT
clamr(2)--COMPLETED
```

arhall0:
I don't know if there is a `TIMELIMIT` job state: https://slurm.schedmd.com/squeue.html#SECTION_JOB-STATE-CODES

Perhaps we should verify if this exists and if not remove the code that mentions it.

pagrubel (approved):
This looks good! Thank you!
I ran both clamr examples with short times. 
No checkpointing results in failed dependency. 
And with checkpointing schedules jobs until one passes and the workflow completes.
Good work!

Issue #1037, opened 2025-04-08
Cleaner Reset/Stop/Start for beeflow core


rstyd:
Sometimes when you go to kill beeflow processes they don't all stop cleanly. We should add code to the component manager within `beeflow/client/core.py` that iterates through the PIDs for all components to check if they all die within some timeout window.

When we go to remove the bee_workdir, we need to confirm that it's empty instead of just throwing an exception and add a timeout for that as well since a slow filesystem makes the amount of time that takes variable. 

When starting, we need to confirm that the PIDs for each component from the previous run are dead.

PR #1039, opened 2025-04-10
Cleaner beeflow core stop and reset

arhall0:
This PR aims to partially address #1037 

This adjusts the `ComponentManager` kill to more effectively kill the `beeflow` components and also uses `Popen.wait` to ensure the sub-components terminate.

File clean-up on reset has been improved, but still seems to leave `neo4j` artifacts due to `.nfs` files.

### DIFF
--- a/beeflow/client/core.py
+++ b/beeflow/client/core.py
@@ -46,6 +46,7 @@ class ComponentManager:
         """Construct the component manager."""
         self.components = {}
         self.procs = {}
+        self.order = []
 
     def component(self, name, deps=None):
         """Return a decorator function to be called."""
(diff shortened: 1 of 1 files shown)


beeflow/client/core.py
```diff
@@ -113,8 +115,8 @@ def poll(self):
                 continue
             returncode = self.procs[name].poll()
             if returncode is not None:
-                log = paths.log_fname(name)
-                print(f'Component "{name}" failed, check log "{log}"')
+                log_fname = paths.log_fname(name)
```
arhall0, line 118: There are some corrections of `pylint` not liking `log` being used to mean multiple things.

beeflow/client/core.py
```diff
@@ -134,9 +136,17 @@ def status(self):
...
         """Kill all components."""
-        for name, proc in self.procs.items():
-            print(f'Killing {name}')
-            proc.terminate()
+        # killing in this order prevents killing celery deps before celery
+        for name in sorted(self.procs.keys()):
+            proc = self.procs[name]
+            # https://slurm.schedmd.com/slurmrestd.html#SECTION_SIGNALS
+            if name in ['slurmrestd']:
+                log.info(f'Sending {name} SIGINT')
```
arhall0, line 144: Previously this function had print statements that weren't seen. There are still some others in the `ComponentManager` but I didn't change those.

The logs are deleted and therefore can't be seen during a `reset` unless something goes wrong, but can be seen for `stop`.

beeflow/client/core.py
```diff
@@ -134,9 +136,17 @@ def status(self):
 
     def kill(self):
         """Kill all components."""
-        for name, proc in self.procs.items():
-            print(f'Killing {name}')
-            proc.terminate()
+        # killing in this order prevents killing celery deps before celery
```
arhall0: There was an error in the `celery` logs about this because `redis` was being shut down first. I don't think this matters much, but this order prevents the error.

beeflow/client/core.py
arhall0: `slurmrestd` doesn't shut down properly from `SIGTERM` for some reason.

PR #1051, opened 2025-04-16
Fail dependent tasks for CANCELLED tasks

pagrubel:


arhall0 (approved):
I tested this with the `clamr-ffmpeg-build` and `clamr-wf-checkpoint` workflows and everything looked as expected. The diff is pretty minimal so I don't foresee many possible issues.

For `clamr-ffmpeg-build` I used scancel on the `clamr` job as soon as it was running and it ended in the state:

```shell
Archived/Failed
clamr--CANCELLED
ffmpeg--DEP_FAIL
```

I ran the same test for `clamr-wf-checkpoint` with the same result. I also checked if cancelling one of the restart jobs made any difference but this too worked as expected and ended in the final state:

```shell
Archived/Failed
clamr--TIMEOUT
ffmpeg--DEP_FAIL
clamr(1)--CANCELLED
```

Issue #1049, opened 2025-04-16
Contents of pre_script are repeated in Slurm script


jthrussell:
I have noticed that whenever I specify a pre_script through the API the resultant slurm script will have the contents of the pre_script repeated several times. I haven't done thorough testing but it seems like the contents are repeated typically 2-4 times, I don't know why there is a variation.

cmbiwer:
I would say this is high priority.

cmbiwer:
It looks like maybe ``post_script`` is unaffected though.

cmbiwer:
High priority is that we need to be able to view the stdout/stderr for post/pre.

pagrubel:
@rstyd Was this solved in PR #1052

jthrussell:
This issue has been resolved.

Issue #1046, opened 2025-04-15
Setting QOS, time limit, etc. with BEE API using Slurm doesn't seem to work


cmbiwer:
Doing `` Slurm(time_limit="00:44:00")`` doesn't seem to be setting the time limit in the sbatch job.

It looks similar for other arguments like QOS as well.

pagrubel:
Yes, that is broken and is in the works. The requirements work for BEE; the api is broken.

cmbiwer:
We can talk about it, but this is probably the most pressing thing.

pagrubel:
This has been fixed in PR #1052

cmbiwer:
I'll test this out in the afternoon. Thanks!

PR #1052, opened 2025-04-22
Fix CWL API and Move stdout/stderr

rstyd:
This PR addresses  #1046, #1049, and additionally moves the stderr and stdout for tasks to the workdir.

### DIFF
--- a/beeflow/common/cwl/cwl.py
+++ b/beeflow/common/cwl/cwl.py
@@ -374,7 +374,7 @@ class SlurmRequirement:
         if self.partition:
             sched_dump['beeflow:SlurmRequirement']['partition'] = self.partition
         if self.qos:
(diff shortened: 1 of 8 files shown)


pagrubel:
This seems to be working, although it is leaving extra output files around with our examples. I tried and sample python script that used the API and that worked well. It included the qos & reservation and partition flags.

Is there an example that adds Checkpoint/Restart or is that not available in this PR? 
Can you fix the unit tests? and the linter?

pagrubel (approved):
Good work!

This looks good and works well.
I tested this all out with several workflows and generated cwl using the python api. 
Time limit and qos work.
Nice to see the stderr, stdout and script in the workdir
We might want to in the future put some of them in a subdirectory, but that can be another issue.

PR #1054, opened 2025-04-24
Add some basic regression testing for `beeflow/common/cwl`

arhall0:
Starting to add some basic regression tests for CWL which currently has 0 line coverage. Some of these will be incredibly simplistic.


beeflow/common/cwl/cwl.py
```diff
@@ -321,7 +321,7 @@ def dump(self):
         if self.container_name:
             docker_dump['DockerRequirement']['beeflow:containerName'] = self.container_name
         if self.force_type:
-            docker_dump['DockerRequirement']['beeflow:force=Type'] = self.force_type
+            docker_dump['DockerRequirement']['beeflow:forceType'] = self.force_type
```
arhall0, line 324: This looked like a typo.

beeflow/common/cwl/cwl.py
```diff
@@ -406,8 +406,8 @@ def dump(self):
         checkpoint_dump[req_name]['container_path'] = self.container_path
         checkpoint_dump[req_name]['file_regex'] = self.file_regex
         checkpoint_dump[req_name]['restart_parameters'] = self.restart_parameters
-        checkpoint_dump[req_name]['num_tries'] = self.enabled
-        return req_name
+        checkpoint_dump[req_name]['num_tries'] = self.num_tries
+        return checkpoint_dump
```
arhall0, line 410: I think this was a bug.

beeflow/common/cwl/cwl.py
```diff
@@ -627,6 +627,4 @@ def dump_inputs(self, path=None):
 
     def __repr__(self):
         """Return CWL file as a string."""
-        stream = StringIO()
-        yaml.dump(self.dump_wf(), stream)
-        return stream.getvalue()
+        return self.dump_wf()
```
arhall0, line 630: The above is already done for `dump_wf`

beeflow/common/cwl/examples/comd/comd.cwl
pagrubel, line 1: Is this file necessary? Isn't it the output running the python script to generate it? I was wondering since we have various cwl files in many places and maybe it was added when trying the scripts out.

beeflow/common/cwl/examples/comd/comd.yml
pagrubel, line 1: Same comment as for the cwl file.
Is this file necessary? Isn't it the output running the python script to generate it? I was wondering since we have various cwl files in many places and maybe it was added when trying the scripts out.

beeflow/common/cwl/examples/cat-grep-tar/cat-grep-tar.cwl
pagrubel, line 1: Is this file necessary? Isn't it the output running the python script to generate it? I was wondering since we have various cwl files in many places and maybe it was added when trying the scripts out.

pagrubel (changes requested):
Should there be an ignore for the generated cwl and yml files?

pagrubel:
Thanks this looks really good. Just let me know what you think about my comments on the cwl and yml files.

arhall0:
> Should there be an ignore for the generated cwl and yml files?

The `.cwl` and `.yml` files were added to be used in regression testing in `beeflow/tests/test_cwl_workflow.py`. We could store the text in a variable in that file if you don't want to more files to be added. Or move them to the `tests` folder to be more clear of their use.

pagrubel:
I think they should be moved to the tests folder. That will indicate that they are used for tests.

pagrubel:
> > Should there be an ignore for the generated cwl and yml files?
> 
> The `.cwl` and `.yml` files were added to be used in regression testing in `beeflow/tests/test_cwl_workflow.py`. We could store the text in a variable in that file if you don't want to more files to be added. Or move them to the `tests` folder to be more clear of their use.

I think they should be in the tests folder probably a subfolder in tests like tests/cwl_files or some such thing.

pagrubel (approved):
Thank you for adding these tests, they look great!

PR #1057, opened 2025-04-28
Disable builder when no DockerRequirement present

kchilleri:
The builder only runs if there is a DockerRequirement in the CWL, but the worker (in ```worker.py```) calls the ```ContainerRuntimeInterface``` to load the container runtime drivers even if a container is not required. Because of this, the crt logs are still appearing in the ```task_manager.log``` file. We could probably make this a separate issue.

### DIFF
--- a/beeflow/common/build/container_drivers.py
+++ b/beeflow/common/build/container_drivers.py
@@ -82,7 +82,7 @@ class CharliecloudBuildDriver(ContainerBuildDriver):
                 docker_requirements = docker_requirements.union(hint_docker_requirements)
                 hint_str = f'{set(hint_docker_requirements)}'
                 log.info(f'task {self.task.name} {self.task.id} hint '
-                         'DockerRequirements: {hint_str}')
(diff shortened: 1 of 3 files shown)


kabir-vats (approved):
These changes run and meet all of the requirements that I see. No issues.

pagrubel (dismissed):
I still see some logs for steps that do not have DockerRequirements

```
[2025-07-02 15:59:38,130] beeflow.task_manager.background:submit_task(): Job Submitted 'grep0' job_id: 16561765 job_state: PENDING
[2025-07-02 15:59:38,272] beeflow.common.crt.charliecloud_driver:run_text(): Build container archive directory is: /vast/home/pagrubel/.beeflow/container_archive
[2025-07-02 15:59:38,272] beeflow.common.crt.charliecloud_driver:run_text(): No beeflow:containerName provided.
```
I suppose this is not really part of the builder, however, we shouldn't be accessing the container runtime if there is no need for a container.

Also, we should add tests for the checks this drops coverage.

PR #1059, opened 2025-05-01
Add CheckpointRequirement to Python CWL API

arhall0:
Adds support for `CheckpointRequirement` to the CWL generation. Aspects of this were already written.

I added a test and an example using a simple `clamr` workflow based on `ci/clamr-wf-checkpoint`; but I removed the `ffmpeg` step for simplicity.

If you want to test the workflow that is generated:

```shell
cd beeflow/common/cwl/examples
mkdir clamr
python3 clamr_checkpoint.py
cp ../../../../ci/test_workflows/clamr-wf-checkpoint/Dockerfile.clamr-ffmpeg .
beeflow submit test clamr clamr/clamr.cwl clamr/clamr.yml .
```

Side note: this workflow uses SlurmRequirement to set a time limit of 10 seconds, but this may be smaller than what is allowed. For example, it can get rounded to the nearest minute depending on the slurm configuration where its run.

### DIFF
--- a/beeflow/common/cwl/cwl.py
+++ b/beeflow/common/cwl/cwl.py
@@ -359,7 +359,7 @@ class SlurmRequirement:
     """Represents a beeflow custom MPI requirement."""
 
     account: str = None
-    time_limit: int = None
+    time_limit: str = None
     partition: str = None
     qos: str = None
     reservation: str = None
--- a/beeflow/common/cwl/cwl.py
+++ b/beeflow/common/cwl/cwl.py
@@ -72,6 +72,7 @@ class InputBinding:
 
     prefix: str = None
     position: int = None
+    value_from: str = None
 
     def dump(self):
         """Dump returns dictionary that will be used by pyyaml dump."""
@@ -80,6 +81,8 @@ class InputBinding:
             binding_yaml['inputBinding']['position'] = self.position
         if self.prefix:
             binding_yaml['inputBinding']['prefix'] = self.prefix
+        if self.value_from:
+            binding_yaml['inputBinding']['valueFrom'] = self.value_from
         return binding_yaml
 
     def __repr__(self):
--- a/beeflow/common/cwl/examples/clamr_checkpoint.py
+++ b/beeflow/common/cwl/examples/clamr_checkpoint.py
@@ -1,0 +1,66 @@
+"""COMD driver for CWL generator."""
+
+import pathlib
+from beeflow.common.cwl.workflow import (
+    Task,
+    Input,
+    Output,
+    MPI,
+    Charliecloud,
+    Workflow,
+    Slurm,
+    Script,
+    Checkpoint,
+)
+
+
+def main():
+    clamr_task = Task(
+        name="clamr",
+        base_command="/CLAMR/clamr_cpuonly",
+        stdout="clamr.txt",
+        stderr="clamr.err",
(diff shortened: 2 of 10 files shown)


beeflow/common/cwl/cwl.py
```diff
@@ -359,7 +359,7 @@ class SlurmRequirement:
     """Represents a beeflow custom MPI requirement."""
 
     account: str = None
-    time_limit: int = None
+    time_limit: str = None
```
arhall0, line 365: `sbatch` can accept a lot of different formats (https://slurm.schedmd.com/sbatch.html). Only `int` would be limited to minutes.

pagrubel:
> Adds support for `CheckpointRequirement` to the CWL generation. Aspects of this were already written.
> 
> I added a test and an example using a simple `clamr` workflow based on `ci/clamr-wf-checkpoint`; but I removed the `ffmpeg` step for simplicity.
> 
> If you want to test the workflow that is generated:
> 
> ```shell
> cd beeflow/common/cwl/examples
> mkdir clamr
> python3 clamr_checkpoint.py
> cp ../../../../ci/test_workflows/clamr-wf-checkpoint/Dockerfile.clamr-ffmpeg .
> beeflow submit test clamr clamr/clamr.cwl clamr/clamr.yml .
> ```
> 
> Side note: this workflow uses SlurmRequirement to set a time limit of 10 seconds, but this may be smaller than what is allowed. For example, it can get rounded to the nearest minute depending on the slurm configuration where its run.

It might be useful to put this information in a Readme somewhere, eventually maybe in the documentation for the cwl which doesn't exist yet.

pagrubel (changes requested):
@arhall0 This looks good! I would like to refer users to the example and would like to have the ffmpeg step in that example. Would you put one in the upper level examples directory with the instructions to run it. Having the dependent task shows them a better example since I think they will be running multiple simulations then doing a post-processing. Thanks

pagrubel:
> @arhall0 This looks good! I would like to refer users to the example and would like to have the ffmpeg step in that example. Would you put one in the upper level examples directory with the instructions to run it. Having the dependent task shows them a better example since I think they will be running multiple simulations then doing a post-processing. Thanks

Actually it would probably be good to only put the python script that generates the cwl and the Dockerfile in the examples directory, so they can go through the entire process without already having the cwl.

beeflow/common/cwl/cwl.py
```diff
@@ -72,6 +72,7 @@ class InputBinding:
 
     prefix: str = None
     position: int = None
+    value_from: str = None
```
arhall0, line 75: This was needed for `ffmpeg` but it wasn't yet supported.

pagrubel:
This looks good. I ran it and added the DockerFile to make it easy to find. I'd like @rstyd  to verify all will work well with the other framework.
@arhall0 Thank you

arhall0:
> This looks good. I ran it and added the DockerFile to make it easy to find. I'd like @rstyd to verify all will work well with the other framework. @arhall0 Thank you

Thanks! I had moved the DockerFile there, but then forgot to add it with git.

rstyd (approved):
Hey Aaron, This looks great! Approved.

pagrubel (approved):

pagrubel:
Thank you, @arhall0

PR #1062, opened 2025-05-12
Adding tests to improve `bee_client` coverage

arhall0:
`bee_client` has fairly low coverage, so I will add some simple regression tests for many of the functions.

### DIFF
--- a/beeflow/client/bee_client.py
+++ b/beeflow/client/bee_client.py
@@ -558,8 +558,6 @@ def list_workflows():
     else:
         typer.echo("There are currently no workflows.")
 
-    logging.info('List workflows:  {resp.text}')
-
 
 @app.command()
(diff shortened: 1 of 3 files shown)


arhall0:
`bee_client` is up to 41% coverage. I'll continue writing tests for it in a follow-up PR.

beeflow/client/bee_client.py
```diff
@@ -558,8 +557,6 @@ def list_workflows():
     else:
         typer.echo("There are currently no workflows.")
 
-    logging.info('List workflows:  {resp.text}')
```
arhall0, line 561: This function doesn't have `resp.text` anymore; it was refactored.

pagrubel (approved):
This looks good, thank you for adding the coverage!

PR #1063, opened 2025-05-16
More testing of `bee_client`

arhall0:
Continuing from #1062

### DIFF
--- a/beeflow/client/bee_client.py
+++ b/beeflow/client/bee_client.py
@@ -274,10 +274,7 @@ def match_short_id(wf_id):
                          "workflow ID {matched_ids[0]}")
             long_wf_id = matched_ids[0]
             return long_wf_id
-    else:
-        sys.exit("There are currently no workflows.")
-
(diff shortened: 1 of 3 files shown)


beeflow/client/bee_client.py
```diff
@@ -274,10 +274,7 @@ def match_short_id(wf_id):
                          "workflow ID {matched_ids[0]}")
             long_wf_id = matched_ids[0]
             return long_wf_id
-    else:
-        sys.exit("There are currently no workflows.")
-
-    return None
+    sys.exit("There are currently no workflows.")
```
arhall0, line 277: This change was to satisfy `pylint` but also prevent an unreachable line from lowering coverage.

arhall0:
There are a couple functions that still need coverage in `bee_client`. I will work on them in a follow-up PR.

pagrubel (approved):
Thank you so much. This gets us closer to our goal for coverage.

PR #1070, opened 2025-05-21
Fix blank restart_parameters shows None

### LINKED ISSUE #1066: If the Restart parameter is left blank None shows up in the command for running the restart

arhall0:
Closes #1066 

This adds a None check on parsing of hint parameters which prevents the 'None' from showing up downstream when `restart_parameters` is blank. 

I added a simple regression test which should fail prior to this patch.

Additionally, I checked the `*.sh` files created when running `ci/test_workflows/clamr-wf-checkpoint` after setting `restart_parameters` to nothing. Prior to the patch 'None' showed up in the command, but after the patch it does not. The workflow errors because the command is incorrect without the `-R`, but that is expected.

### DIFF
--- a/beeflow/common/parser/parser.py
+++ b/beeflow/common/parser/parser.py
@@ -339,7 +339,7 @@ class CwlParser:
             for req in requirements:
                 items = {}
                 for k, v in req.items():
-                    if k != 'class':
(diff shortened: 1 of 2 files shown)


pagrubel (approved):
This looks good, thanks for adding the test.

PR #1071, opened 2025-05-22
Fix restarts script file names surrounded by quotes

### LINKED ISSUE #1067: For restarts the script file names are surrounded by quotes
For example, the new steps that we generate like the following: clamr-dca3.sh 'clamr(1)-ceea.sh'

arhall0:
Closes #1067 

This PR implements the suggestion in #1067 to change restart name convention to `file_name-1.sh`, `file_name-2.sh`, etc from the original `file_name(1).sh` which resulted in files having quotes (`'file_name(1).sh'`) when examining via `ls` or similar.

This change was already covered by a test, so I just updated it to match the new convention.

### DIFF
--- a/beeflow/common/wf_interface.py
+++ b/beeflow/common/wf_interface.py
@@ -121,12 +121,12 @@ class WorkflowInterface:
         # Pattern match on task name
         # Append (1) if not in name already
         # Increment number on each restart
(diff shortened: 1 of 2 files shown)


pagrubel (approved):
Thank you this is perfect!

PR #1072, opened 2025-05-23
Support additional parameters when restarting

### LINKED ISSUE #1068: Add parameters to generalize the restart requirement
We could have add strings with positions supplied. This needs to be done for the API as well.

An alternative is that we let the user have several restart parameters associated with their own file names or strings
so the command would end up being something like:

srun --nodes=1 command restart-parameter1 filename1 restart-parameter2 filename2 ...

arhall0:
Closes #1068

This adds a new option to `beeflow:CheckpointRequirement` called `add_parameters`. Anything specified will be appended to the end of the run command on restart. The CWL API has been updated to implement this.

### DIFF
--- a/beeflow/common/cwl/cwl.py
+++ b/beeflow/common/cwl/cwl.py
@@ -1,6 +1,7 @@
 """Create and manage CWL files."""
 from dataclasses import dataclass
 from io import StringIO
+from typing import Optional
 import ruamel.yaml
 
 # Create the global object for ruamel.ymal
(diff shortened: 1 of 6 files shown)


arhall0:
Likely this should be merged after #1070 as the issue affecting `restart_parameters` will also affect `add_parameters` until that is merged.

pagrubel (approved):
Thank you, this looks good and I tested it to make sure it adds the parameters.

Issue #1060, opened 2025-05-01
Workdir subdirectories for task scripts and possible other files


pagrubel:
The scripts for all the tasks and other task information go in the workdir for the workflow. This will become quite cluttered when a workflow has potentially thousands of tasks. We need to make subdirectories to put BEE files in the workdir, such as the batch scripts that were run for a task. This might also be a good place to put other files like more information from slurm on the job. Perhaps this information could eventually be also copied to the archive for the workflow.

By default the subdirectory can be named task.name-<first 4 numbers in task-id>

But we need to give the user an opportunity to name the subdirectory

sbzeytun:
I am working on this.

PR #1074, opened 2025-05-28
Adds CWL option to change working directory for task

arhall0:
I probably won't have a chance to completely finish this, so adding it for reference. 

This change allows you to specify a different working directory for a task:

```yaml
hints:
  beeflow:TaskRequirement:
    workdir: ~/BEE_jobs
```

At the moment, this doesn't create the directory, just uses it. Additionally this is using the current directory the submit command is run from as the relative path, but this should probably be relative to the working directory specified by the submit command.


arhall0:
> Additionally this is using the current directory the submit command is run from as the relative path, but this should probably be relative to the working directory specified by the submit command.

I corrected this in the most recent commit. If the task workdir is absolute it will use it as is, otherwise it will assume it is relative to the directory specified by the submit command. This corrected the integration test failures.

The current unit test failures are from `test_parser.py` which has regression tests on things I changed to implement this. These are not easily updated because I changed the `glob` of output files to be absolute, which would include different paths at test time; so the way this is checked will need to be altered.

Additionally, it would be good to add some new tests for some of the extra things I added, like `resolve_step_workdir`.

Another thing to consider is where to store the `.sh` and `stdout` and `stderr` files. Right now we are considering putting them in their own folders per #1060. I think this could work well with this PR by instead putting these files in a subdir of the task workdir that is created at runtime; so something like `task.name-<first 4 numbers in task-id>_submit_files`; a slight variation of what was suggested in #1060. This would allow for the desired behavior in that issue, but also work well with this PR. I think #1052 was written in a way that this behavior would be easy to implement.

pagrubel:
Thank you, I agree if we make a directory for the .sh, stderr and stdout files it should go in this subdirectory.

arhall0:
Additionally, this PR does not currently include implementation of this functionality for the CWL API.

arhall0:
I was able to do a lot of the things I mentioned. So maybe this is ready for review.

beeflow/common/cwl/workflow.py
```diff
@@ -137,6 +137,15 @@ def requirement(self):
                                      enabled=self.enabled)
 
 
+@dataclass
+class TaskReq(TaskRequirement):
+    """Get Checkpoint Requirements."""
```
pagrubel:
```suggestion
    """Get Task Requirements."""
```

pagrubel:
If the directory does not exist, the tasks ends in a SUBMIT_FAIL because it is attempting to write the script in a directory that doesn't exist. An easy fix is to create the directory if it does not exist.

pagrubel:
This seemed to work well until I used a container, specifically with the clamr example. I've did make some changes but one of the problems is the location for the ffmpeg inputs is not relative and because of that it does not translate to the binding in the container. Overnight tests fail due to the same problem.

pagrubel:
One solution may be to use a different path when running in a container. Here is the alternate path that is used. I  need to figure out how to translate this to a subsequent task though. Just documenting it here:
In Charliecloud driver:  
``task.workdir is /vast/home/<username>/beeworkdir/clamr-ffmpeg-build``
``ctr_workdir_path is /home/<username>/beeworkdir/clamr-ffmpeg-build ``

pagrubel:
@rstyd This is ready for review. Container runs now work and the overnight tests with checkpoint/restart pass.

Leahh02 (approved):
Looks good to me. I got it to work with an altered version of cat-grep-tar and clamr. I think it might be helpful to have an example of this in our examples or data directories, but that doesn't have to be in this PR.

PR #1077, opened 2025-06-03
Create subdirectories for task scripts, outputs, and errors #1060

### LINKED ISSUE #1060: Workdir subdirectories for task scripts and possible other files

sbzeytun:
Closes #1060 

This will allow workdir to be less cluttered when workflow runs many tasks.


pagrubel:
This looks great. The archives for some reason have been getting all the outputs we should only be archiving what is in the task-shortid subdirectories created by this PR now.

pagrubel:
@sbzeytun The problems with the archives are not caused by this PR, however just copy the new subdirectory created by this PR into the archive instead of each individual file. I'll submit an issue for the other problems.

pagrubel (approved):
This looks great and I tested it with out checkpoint/restart example. Clears up the clutter

pagrubel:
It's also a good example of getting rid of the lazy formatting in logging.

PR #1080, opened 2025-06-17
Relax pylint version

pagrubel:


Leahh02 (commented):
Looks good! My only question is if we want to change `pip install pylint==3.2.7 2>&1 >/dev/null` to `pip install pylint>=3.2.7 2>&1 >/dev/null` in .github/workflows/pylint.yml

pagrubel:
Good suggestion. I'll make that change

Leahh02 (approved):
Looks good!

Issue #1081, opened 2025-06-17
Check Graphviz Availability


Leahh02:
Check if graphviz is available on the system. If not, make a log and make it so dags aren't made -- disable the dag creation that is done when the workflow completes, and probably create a little error message if the user tries to do `beeflow dag $wfid`

Leahh02:
@pagrubel The process of making the DAG involves exporting a graphml file from neo4j and then using graphviz to turn that grpahml file into a DAG. We keep both the graphml files and the DAGs created during the workflow. Do we still want to export the graphml file if the system doesn't have graphviz? 

I guess an example of where that could be useful is if someone was running a workflow on a system without graphviz they could then take the graphml file and turn it into  a DAG on a different system if they wanted to later.

pagrubel:
@Leahh02 good point I think you should keep the graphml file. Maybe we could add a utility for conversion, if that's appropriate.

Leahh02:
> [@Leahh02](https://github.com/Leahh02) good point I think you should keep the graphml file. Maybe we could add a utility for conversion, if that's appropriate.

I think that would be helpful. I'll make an issue for that.

pagrubel:
Satisfied by #1083

PR #1083, opened 2025-06-19
Issue1081/check graphviz avail

Leahh02:
This PR addresses Issue #1081. During beeflow startup there's a check to see if graphviz is on the system. If it's not this gets logged. If someone tries to run `beeflow dag $wf_id $output_dir` without having graphviz, the graphml will be saved to the workflow directory in .beeflow, like normal, and it'll be saved to the output_dir where the dags would typically get put. I've also added a test file for wf_utils. My assumption is that more tests will get added to that file over time. We can choose whether or not to keep it -- this was added in the last commit so it should be easy to deal with.


pagrubel:
This looks good! I cannot test without dot though.

pagrubel (approved):
Looks good. I cannot try it on a system without graphviz, but all works well and thanks for the tests.

PR #1087, opened 2025-07-02
Provide more information when querying a workflow 

sbzeytun:
Provides more information of each task in a workflow, such as the start time, time left, and the job name, when using the 'beeflow query' command.


pagrubel:
I think including the job id and a shortened job name be useful information.  I'm not sure why there are three dictionaries in metadata.json and what time_left means once a task has been completed.
Can you update the documentation to include what you are storing and what is on the screen for the query

sbzeytun:
The three dictionaries stored in metadata.json represent periodic updates to the task information. This file is overwritten each time the metadata is updated or changed. I can change this so it only includes the most recent updates. I feel like having access to the duration/time_left of a task could be useful for users interested in seeing how long each task took, since some tasks depend on each other, and optimizing their workflows.  However, I can change these fields based on what's more important. I will also shorten the task names and update the documentation.

pagrubel:
This work was done in a different PR. I am closing it.

PR #1089, opened 2025-07-02
Support no inputs

### LINKED ISSUE #1035: bee CWL API tasks should be able to have 0 inputs
We need to support tasks with no inputs in the CWL API. I think this will also require changes to the parser or workflow interface.

### LINKED ISSUE #1044: BEE API requires an Input() in order for workflow to run
We have to specify a ``Input()`` for a ``Task``, otherwise the workflow will not submit. It will stay in the ``WAITING`` stage if you check the jobs with ``beeflow query``.

At the very least, it should give an error or something. But preferably it wouldn't need an ``Input()``.

kabir-vats:
Changes neo4j cypher and cwl parser to allow workflows and tasks with no inputs.

### DIFF
--- a/beeflow/common/gdb/neo4j_cypher.py
+++ b/beeflow/common/gdb/neo4j_cypher.py
@@ -239,8 +239,8 @@ def add_dependencies(tx, task, old_task=None, restarted_task=False):
         tx.run(restarted_query, old_task_id=old_task.id, new_task_id=task.id)
         tx.run(dependency_query, task_id=task.id)
     else:
-        begins_query = ("MATCH (s:Task {id: $task_id}) "
-                        "MATCH (w:Workflow (id: s.workflow_id}) "
+        task_of_query = ("MATCH (s:Task {id: $task_id}) "
+                        "MATCH (w:Workflow {id: s.workflow_id}) "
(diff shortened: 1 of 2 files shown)


beeflow/common/gdb/neo4j_cypher.py
```diff
@@ -239,12 +239,9 @@ def add_dependencies(tx, task, old_task=None, restarted_task=False):
         tx.run(restarted_query, old_task_id=old_task.id, new_task_id=task.id)
         tx.run(dependency_query, task_id=task.id)
     else:
-        begins_query = ("MATCH (s:Task {id: $task_id})<-[:INPUT_OF]-(i:Input) "
-                        "WITH s, collect(i.source) AS sources "
-                        "MATCH (w:Workflow {id: s.workflow_id})<-[:INPUT_OF]-(i:Input) "
-                        "WITH s, w, sources, collect(i.id) AS inputs "
-                        "WHERE any(input IN sources WHERE input IN inputs) "
-                        "MERGE (s)-[:BEGINS]->(w)")
+        task_of_query = ("MATCH (s:Task {id: $task_id}) "
```
rstyd, line 242: Hey Kabir can you give a bit more info on what this change does?

PR #1091, opened 2025-07-03
Use Pydantic for Object Models

### LINKED ISSUE #1094: Use Pydantic for Object Models and Endpoint Validation
Pydantic allows us to verify json schemas at endpoint and make sure that internal communication follows protocols (kind of like protobuffs).

Using pydantic at flask endpoints is also a prerequisite for transitioning to FastAPI

kabir-vats:
This change swaps all of the custom models to pydantic models for type safety and as a precursor to a change to FastAPI.

BEE has a lot of interfaces that communicate with each other. They were previously just communicating in raw JSON format. So, JSON format was specified on one side, the other side will pickle a message and send it over to be unpickled. This is not good practice; if one interface’s behavior changes, the other interface would have to have its behavior changed too. Also, many bugs go undetected through the pickle-unpickle process. Pydantic solves this issue. With pydantic, protocols to communicate with interfaces are specified as python objects, so the interfaces can always know the format they are accepting, and client can always know what format to make a request in.
 
Additionally, internal requests and responses over HTTP are now all serialized as Pydantic objects. 

Changelog:
- Changes the models for Tasks, Workflows, Hints, Requirements, etc. to be backed by Pydantic models
- Changes the interfaces between all of the REST APIs (workflow manager, task manager, scheduler, bee client which calls these APIs) to use pydantic models as a "protobuff" style object in communication. This prevents future communication bugs.
- Updates declarations of the refactored objects. Main change there is that pydantic prefers keyword assignment as opposed to positional assignment, which is largely good (it makes it a lot easier to spot bugs, there were a couple existing object declarations that were actually incorrect due to miscounting positional arguments)
- In general, makes communication with the endpoints a lot more consistent and standardized. This makes a future migration from Flask to FastAPI a lot less of a leap.


beeflow/wf_manager/resources/wf_utils.py
```diff
@@ -17,6 +16,12 @@
 from beeflow.common import paths
 from beeflow.common.db import wfm_db
 from beeflow.common.db.bdb import connect_db
+from beeflow.scheduler.models import (
+    ScheduleTasksRequest,
+    ScheduleTasksResponse,
+    SchedulerTask,
+)
+from beeflow.task_manager.models import SubmitTasksRequest
```
rstyd, line 24: This request appears to be getting imported twice.

beeflow/wf_manager/resources/wf_utils.py
```diff
@@ -242,16 +245,19 @@ def submit_tasks_scheduler(tasks):
...
-        log.error('Unable to connect to scheduler to submit tasks.')
+        log.error("Unable to connect to scheduler to submit tasks.")
         return "Did not work"
 
     if resp.status_code != 200:
-        log.info("Something bad happened %s",resp.status_code)
+        log.info("Something bad happened %s", resp.status_code)
         return "Did not work"
-    return resp.json()
+    return ScheduleTasksResponse.model_validate(resp.json()).tasks
```
rstyd, line 260: Could this call throw an exception if the model isn't valid?

kabir-vats: Yes and it should; it would indicate that the json format within the scheduler has changed but it the model hasn't been updated, so workflow manager won't know how to handle the response.

pagrubel (approved):
This is great! I also ran the overnight, local tests on this and they pass.

PR #1095, opened 2025-07-08
Update the GDB Workflow Status on Archive

### LINKED ISSUE #740: Pausing or Resuming a Stopped Workflow
Currently, if we try to pause or resume a stopped workflow it causes an error because the REST endpoint relies on wfi.get_workflow_state which only works as long as the workflow database is up. We should change this code to use the wfm db instead.

### LINKED ISSUE #1038: Pausing an archived workflow
Right now you can pause an archived workflow if the "delete_completed_workflow_dirs" config option is set to true. If it's set to false you get an error log to report to the bee team, which is caused by the line `wf_utils.update_wf_status(wf_id, 'Paused')` in wf_actions.py because it's trying to write to the bee_wf_status file in the workflow directory. **Really what should be happening is the message `Cannot pause workflow. It is currently Archived(/Failed...).` should be printed.**

**Why is this happening:**
In wf_actions.py, the variable wf_state is assigned by checking the status in neo4j, and then the check `if option == 'pause' and wf_state in ('RUNNING', 'INITIALIZING'):` is run. Right now when a workflow completes or is archived, the neo4j workflow status doesn't get updated, it just stays RUNNING.

If the wf_state was coming from the wfm.db I don't think this would be a problem. This also wouldn't be a problem if we were assiging wf_state from neo4j, but the neo4j workflow status got updated.

The neo4j workflow status can be found by going into the workflow directory or the archive, and looking at the most recent graphml file in the graphmls directory. Then look for the workflow node and find it's state. For example:
`<ns0:node id="n1" labels=":Workflow"><ns0:data key="labels">:Workflow</ns0:data><ns0:data key="reqs">false</ns0:data><ns0:data key="hints">false</ns0:data><ns0:data key="restart">false</ns0:data><ns0:data key="id">2e764d146bca480aaf135a8f44a679f2</ns0:data><ns0:data key="name">workflow</ns0:data><ns0:data key="state">RUNNING</ns0:data></ns0:node>` the state here is RUNNING.

**Side note:** I don't think INITIALIZING is a state that is ever assigned to the workflow status in neo4j, but I could be wrong. I know the workflow state in wfm.db gets set to Initializing.

kabir-vats:
The status in the graph database remains 'running' even when a workflow is archived, causing unintended behavior when users try to interact with the archived workflow. This pull request makes sure to set that state.

Resolves #740

### DIFF
--- a/beeflow/tests/test_wf_update.py
+++ b/beeflow/tests/test_wf_update.py
@@ -3,6 +3,7 @@
 import os
 import pytest
 from beeflow.wf_manager.resources import wf_update
+from beeflow.tests.mocks import MockWFI
 
 
 @pytest.mark.parametrize(
@@ -20,6 +21,12 @@ def test_archive_workflow(tmpdir, mocker, test_function, expected_state):
     mocker.patch("os.path.expanduser", return_value=str(tmpdir))
     mocker.patch(
         "beeflow.wf_manager.resources.wf_utils.get_workflow_dir", return_value=workdir
+    )
+    mock_wfi = MockWFI()
+    mock_wfi.set_workflow_state = mocker.MagicMock()
+    mocker.patch(
+        "beeflow.wf_manager.resources.wf_utils.get_workflow_interface", 
+        return_value=mock_wfi
     )
     mocker.patch(
         "beeflow.common.config_driver.BeeConfig.get",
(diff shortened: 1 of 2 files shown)


kabir-vats:
With respect to #1038 , It should probably be printing out things when people pause / resume workflow but doesn't. I want to wait for #1091 to be merged before addressing this part though. I think we can leave this PR unmerged until that gets pushed and then:
- [ ]  Make sure pause / resume are giving feedback to the command line

kabir-vats:
I'm going to close this PR because it's redundant after working on #1105 -> I will moved the linked issues there.

PR #1098, opened 2025-07-09
Add Proper Wait Monitoring for Neo4j Being Up

### LINKED ISSUE #1096: Workflow stuck in Initializing
This occurs when beeflow is started and a workflow is submitted before everything is really running.
The status of the daemon's and associated files should be assessed before being able to submit or before giving access back to the user from a beeflow start command.

You can easily duplicate this and look at the celery log to see what is happening.

kabir-vats:
If someone submits a workflow while bee is still initializing, it will keep trying to verify connection for 60 seconds before erroring, rather than just erroring immediately. This fixes most startup issues people currently face, and it should only trigger on the first workflow for most people.

Resolves #1096

### DIFF
--- a/beeflow/common/deps/neo4j_manager.py
+++ b/beeflow/common/deps/neo4j_manager.py
@@ -5,6 +5,13 @@ import shutil
 import re
 import subprocess
 import time
+import logging
+
+from neo4j import GraphDatabase
+from neo4j import exceptions as neo4j_exc
+
+from beeflow.common.gdb.neo4j_driver import (DEFAULT_USER, DEFAULT_PASSWORD, DEFAULT_HOSTNAME,
+                                             DEFAULT_BOLT_PORT, MAX_WAIT_SECS, RETRY_DELAY_SECS)
 
 from beeflow.wf_manager.resources import wf_utils
 from beeflow.common.db import wfm_db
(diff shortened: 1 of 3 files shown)


pagrubel:
@kabir-vats This looks great, but I did find a corner case where we can still get a workflow stuck in Initializing. If you'd like to work on that in another PR or just have an issue someone else takes up I'm prepared to merge this. It happens if you issue ``beeflow core stop`` and immediately a ``beeflow core start``.  This shouldn't happen often, but we should cover it. So I think we should wait when the user stops beeflow until neo4j is actually stopped to give control back to the user. Would you rather that was an issue and another PR?

kabir-vats:
> @kabir-vats This looks great, but I did find a corner case where we can still get a workflow stuck in Initializing. If you'd like to work on that in another PR or just have an issue someone else takes up I'm prepared to merge this. It happens if you issue `beeflow core stop` and immediately a `beeflow core start`. This shouldn't happen often, but we should cover it. So I think we should wait when the user stops beeflow until neo4j is actually stopped to give control back to the user. Would you rather that was an issue and another PR?

Ah, I hadn't considered that. Yes, I think it's different and slightly nontrivial to check if the process is down. I'm actually unsure the criteria (whether it's because the port is still in use, or specific process is still running, etc.) on this, but I don't think it's going to use the same fix, so it can be a separate issue.

pagrubel (approved):
This looks great and solves problems with actions when submitting or resuming workflows before neo4j is up and running. Thank you.

Issue #781, opened 2024-02-13
Ensure BEE is checking all possible job states


jtronge:
We need to make sure that BEE is checking all possible job states and updating the workflow accordingly. This will affect checkpoint-restart as well as how BEE handles workflows that are still running during a DST.

Job states/statuses for different schedulers:

* [Slurm](https://slurm.schedmd.com/squeue.html#SECTION_JOB-STATE-CODES)
* [Flux](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man1/flux-jobs.html#job-status)

Leahh02:
I'm going to refer to this [Slurm job states](https://slurm.schedmd.com/job_state_codes.html) code page since the page that Jake mentioned say to check this page for a full possible list of states.

- In beeflow/common/worker/flux_worker.py we query the job state and then convert the job state from flux into a BEE state. The flux states mapping does contain all of the possible states from the [flux documentation](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man1/flux-jobs.html#job-status). I guess my only question is what if flux adds a new state? I'm assuming that's not likely, but maybe we should have something where if the flux state isn't in the mapping we make one of those bee error logs that tells the user to contact the development team?
- In beeflow/common/worker/flux_worker.py we query the job state using the [get_job](https://flux-framework.readthedocs.io/projects/flux-core/en/stable/python/job_submission.html#get-job) method from flux. When the job id doesn't exist the get_job method returns None. I don't think we handle that at all. Maybe we should be raising a WorkerError if this happens
- I don't know if we're doing anything with LSF but looking at their [job states documentation](https://www.ibm.com/docs/en/spectrum-lsf/10.1.0?topic=execution-about-job-states) it looks like our bee states mapping in beeflow/common/worker/lsf_worker.py is missing WAIT, EXIT, POST_DONE, POST_ERR. Also, similar to my first bullet point, it looks like in the query_job method in beeflow/common/worker/lsf_worker.py we don't have any way to handle a case where lsf returns a state that's not in our bee states mapping.

- It doesn't look like we're checking for the DEADLINE or the SUSPENDED states that are listed in [Slurm job states](https://slurm.schedmd.com/job_state_codes.html).
  - Reading their descriptions it seems like they should maybe be handled the same way that the PREEMPTED state is in beeflow/task_manager/background.py
[...]

PR #1105, opened 2025-07-11
Wf status refactoring

### LINKED ISSUE #740: Pausing or Resuming a Stopped Workflow

### LINKED ISSUE #919: Workflow Status Updated in Multiple Places

### LINKED ISSUE #1102: Task objects passed around unnecessarily
Between workflow manager, workflow interface, neo4j driver and neo4j cypher there are several queries that use task objects when task ids would be sufficient. This sometimes adds unnecessary database calls.

### LINKED ISSUE #1109: Costly Imports in Bee Client
This is separate from the circular import - bee_client.py does imports every time a user puts in a command. This means costly imports can't be done in top levels of modules imported by bee client, because it adds unneeded latency on user requests.

kabir-vats:
Refactor the way workflow and task statuses are stored and updated.
This PR also addresses import statements within the bee client causing latency.

The reasoning for the changes within this PR is: certain databases were updated in certain places, but not others (we could have a workflow set to "Archived" within the sql database, but not within the graph database) because of inconsistent database usage for status queries. This moves all information to one single source of truth (graph database) and moves all queries regarding workflow status to call the same method (get_wf_status, update_wf_status), which will make changes in the future much easier.

Changelog:
- Remove uses of workflow manager database to manage workflows and tasks (move all queries to the GDB)
- Move one-time imports from the main body of certain modules into functions to reduce overhead when running CLI commands. (saves 2-4 seconds per command)
- For consistency, use task ID to query task database for consistency
- Makes users wait when submitting tasks while the GDB is starting up, as opposed to being returned to the command line
- Move use of celery shared_task to the actual submission of tasks to the task manager, as opposed to the addition of tasks to the database. Negligible time increase and prevents race conditions between instant submissions -> queries
- Actually remove workflows that are removed (via user command) from the graph database


Leahh02:
I didn't have time to really review the PR but I do have 2 things to mention:
1. in beeflow/common/states.py we have the workflow states. In there we have RUNNING, for example. I think this PR has the state as Running. I don't entirely know what states.py really does, but I think it should match what we have.
2. I mentioned this is Issue #781, should the patch method of beeflow/wf_manager/resources/wf_actions.py also check if the workflow status is Pending or Starting? To my understanding, with you PR the states go in the order Initializing -> Pending -> Starting -> Running -> Archived. I haven't really had time to look this over, so maybe it should just stay the way it is.

rstyd (approved):
This looks great! Approved!

pagrubel (approved):
Made sure this works well be beeflow is stopped while a step is running and the workflow continues once beeflow is restarted even the next day. Works with slurmrestd.

Nice work, thank you!

PR #1119, opened 2025-07-22
Fix bug in parser.py with workdirs

### LINKED ISSUE #1115: CWL parser bug when using workdir
The is a bug in `parse_step_outputs` in `parser.py` whereby specifying a `workdir` without specifying a `glob` leads to an attempt to join a string and `None`, which obviously fails. Extra logic should be added to guard against this.

jthrussell:
Closes #1115 

If a workdir is specified for a `Task` then `parse_step_outputs` in `parser.py` will try and join the workdir with the value of `glob`. This PR makes the default value of `glob` an empty string so this action succeeds.

### DIFF
--- a/beeflow/common/parser/parser.py
+++ b/beeflow/common/parser/parser.py
@@ -281,7 +281,7 @@ class CwlParser:
                 raise CwlParseError(f"specified step output {out} not produced by CommandLineTool")
 
             output_type = out_map[out].type
-            glob = None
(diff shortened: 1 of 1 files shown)


pagrubel:
Thanks, @jthrussell, as soon as the final test passes, I'll merge this and make a pre-release for you.

pagrubel (approved):

PR #1120, opened 2025-07-23
Add fire workflow

Leahh02:
This PR adds in my fire prediction workflow. It has the full workflow example and the smaller random forest example.

I think this workflow is ready to be added into this data directory. If we want to eventually make it an official example there are a few things I think should be looked at:
- Is there anything superfluous in my dockerfile?
- Is there an easier way to copy files into the container?
- In the script that uses the CWL API I defined common inputs at the top instead of redefining them for each task. Is this the recommended way?
- I'm relying on stdout to show dependent tasks instead of just the files I want to pass.
- Right now I'm giving a very large param grid in the optimization tasks. The param grid can be smaller so it's faster, but I think this shows how BEE is particularity useful if your models take a while to train

Most of these things are talked about in more detail in my [discussion](https://github.com/lanl/BEE/discussions/1118)


rstyd:
This looks fantastic thanks! To address your first question, instead of copying all the files into the container you should be able to just put them in the workflow directory and they'll be bindmounted in when the workflow runs. We'll have to test that though.

Leahh02:
> This looks fantastic thanks! To address your first question, instead of copying all the files into the container you should be able to just put them in the workflow directory and they'll be bindmounted in when the workflow runs. We'll have to test that though.

Oh that makes sense, thank you!

rstyd:
For the inputs, that's kind of an area I haven't thought through well enough. But I'd recommend defining each input variable right above the first task that uses it.

pagrubel:
I think we should merge this and close the PR. The workflow is in a directory that we use for future work. Great Work Leah!

rstyd (approved):
Approved!

PR #1122, opened 2025-07-23
Dump CWL and YAML as strings

rstyd:
This PR modifies the CWL API so that it will properly return the CWL and YAML files as strings when write_yaml or write_cwl is called.

### DIFF
--- a/beeflow/common/cwl/cwl.py
+++ b/beeflow/common/cwl/cwl.py
@@ -157,7 +157,6 @@ class Inputs:
 
 class RunInputs(Inputs):
     """Represents run inputs."""
-    pass
 
 class CWLInputs(Inputs):
     """CWLInputs is used to generate the YAML file."""
--- a/beeflow/common/cwl/cwl.py
+++ b/beeflow/common/cwl/cwl.py
@@ -157,6 +157,7 @@ class Inputs:
 
 class RunInputs(Inputs):
     """Represents run inputs."""
+    pass
 
 class CWLInputs(Inputs):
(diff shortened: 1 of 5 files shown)


pagrubel:
@rstyd Please fix the pylint issue

beeflow/common/cwl/workflow.py
```diff
@@ -203,10 +203,17 @@ def generate_cwl(self):
...
+            self.cwl.dump_wf(path)
+        else:
+            return self.cwl.dump_wf()
 
-    def write_yaml(self, path):
+    def write_yaml(self, path=None):
         """Write the yaml file."""
         self.cwl.dump_inputs(path)
+        if path:
+            self.cwl.dump_wf(path)
```
lilikoi-l: should this be `self.cwl.dump_inputs(path)`?

lilikoi-l:
Would it also be possible to make `dump_wf` not print the string to the terminal if no filepath is entered?

beeflow/common/cwl/workflow.py
rstyd: Yes working on an update now.

beeflow/common/cwl/cwl.py
```diff
@@ -157,6 +157,7 @@ def __repr__(self):
 
 class RunInputs(Inputs):
     """Represents run inputs."""
+    pass
```
pagrubel:
```suggestion
  
```

pagrubel (commented):
Would this change correct the pylint error and work?

pagrubel (approved):

PR #1128, opened 2025-08-11
Create subdirectories for task scripts, outputs, and errors for Flux worker 

sbzeytun:
This fixes the issue of saving the task script, outputs, and errors in the working directory for the Flux worker. Previously, they were not saved under the task name in the working directory.


pagrubel (approved):
Thank you, this looks good.
I did try it while specifying a workdir using TaskRequirement, but apparently that hasn't been implemented in the flux worker. That is a separate issue. I just wanted to make a note of it.

PR #1129, opened 2025-08-13
Provide more information when using 'beeflow query' for Flux

sbzeytun:
Provides more information of each task in a workflow, such as the start time, time left, and the job name, when using the 'beeflow query' command.


sbzeytun:
Notes for future work/issues:
- Yaml doesn’t recognize flux’s job id, so when it is written to metadata.yaml in the working directory, the format is difficult to read
- Map out scheduler attributes with each other, so that the schedulers have consistent attributes. (ie. If Slurm has an attribute 'Reason', does Flux have a similar attribute that's named differently)
-   When the task is pending, display the reason it is pending. (This is a bit difficult because the configuration file determines which attributes to display. Also, flux doesn't seem to have the attribute 'reason' so inconsistencies emerge)
- There seems to be an issue with test_bee_client. When I test it on Darwin the query test passes, however on Github it does not because of the missing partition column. This column should not be missing, so I'm unsure if it's because Github drops the column due to layout issues.

pagrubel:
This looks good there are some upgrades we need to do that @sbzeytun has outline in issue #1130 
I have run this on a flux system and it works well with some of those things mentioned above.
I believe it is in a place where we should merge it. I just need to do a more thorough review.
Thank you for all you have done!

pagrubel:
This PR is failing in our overnight tests. It may not have to do with the actual changes but I will investigate getting that fixed before merging it.

pagrubel (approved):
Looks good and is working well. The information is now there. There are items that can be improved with time especially the differences for flux. Thank you for all the work you did on this.

Issue #1055, opened 2025-04-25
Issues when numpy is in the environment; requires updating networkx version & python


pagrubel:
Using python version 3.12.8 I get these errors which prevent workflows from finalizing and archiving.

```
[2025-04-25 15:34:25,295] beeflow.task_manager.background:update_jobs(): Job Updated 'tar' job_id: 16378966 job_state: COMPLETED
[2025-04-25 15:34:26,218] beeflow.task_manager.background:process_queues(): Traceback (most recent call last):
  File "/vast/home/pagrubel/beeflow-env/lib/python3.12/site-packages/flask/app.py", line 1484, in full_dispatch_request
    rv = self.dispatch_request()
         ^^^^^^^^^^^^^^^^^^^^^^^
  File "/vast/home/pagrubel/beeflow-env/lib/python3.12/site-packages/flask/app.py", line 1469, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/vast/home/pagrubel/beeflow-env/lib/python3.12/site-packages/flask_restful/__init__.py", line 467, in wrapper
    resp = resource(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/vast/home/pagrubel/beeflow-env/lib/python3.12/site-packages/flask/views.py", line 109, in view
    return current_app.ensure_sync(self.dispatch_request)(**kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/vast/home/pagrubel/beeflow-env/lib/python3.12/site-packages/flask_restful/__init__.py", line 582, in dispatch_request
    resp = meth(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/vast/home/pagrubel/BEE/BEE/beeflow/wf_manager/resources/wf_update.py", line 93, in put
    self.update_task_state(state_update, db)
  File "/vast/home/pagrubel/BEE/BEE/beeflow/wf_manager/resources/wf_update.py", line 179, in update_task_state
    self.handle_state_change(state_update, task, wfi, db)
[...]

PR #1142, opened 2025-10-20
Fix networkx issue when numpy is also installed

rstyd:
This PR addresses #1055 by updating the networkx version. This did require tightening the python versions we support, but that doesn't impact any other applications that currently depend on bee.


pagrubel (approved):
Looks good, I did add documentation and tested this with numpy in my environment, the workflows now get archived!

PR #1144, opened 2025-10-27
Enable Slurm Requirement from Inputs

rstyd:
This PR adds a new


pagrubel:
It looks like an MPI or Slurm Requirement can be defined by both an input file and/or a requirement in the cwl. Should we make it one or the other. If there are any duplicates it causes and error.

pagrubel (approved):
This looks good. We just need a safe guard so users don't submit both cwl specifications for the values and input from a file. See Issue #1161

PR #1146, opened 2025-10-30
Impact multiple workflows with the same command 

### LINKED ISSUE #1043: beeflow canceling multiple workflows
If I'm developing a workflow, I often need to cancel/remove everything wholesale. But at the moment this involves calling ``beeflow list``, copying some hash, and then calling more commands.

I think it would be nice to have a ``beeflow cancel --all`` type command. And for multiple workflows ``beeflow cancel WF1 WF2 WF3`` to do several at a time.

Similar with ``beeflow remove``, I had to go through copying them all. I'd like to just remove everything or multiple ones.

There may be other commands but those two I use a lot.

kabir-vats:
Pause, Resume, Cancel, or Remove multiple workflows, either by specifying a list of workflow ids or specifying --all to impact all eligible workflows

### DIFF
--- a/beeflow/client/bee_client.py
+++ b/beeflow/client/bee_client.py
@@ -23,6 +23,7 @@ import textwrap
 import time
 import importlib.metadata
 import importlib
+from typing import List, Optional
 import jsonpickle
 import requests
 import typer
@@ -305,6 +306,19 @@ def match_short_id(wf_id):
             long_wf_id = matched_ids[0]
             return long_wf_id
     sys.exit("There are currently no workflows.")
+
+
+def match_short_ids(wf_ids):
+    """Match user-provided short workflow IDs to full workflow IDs."""
+    long_wf_ids = []
+    for wf_id in wf_ids:
+        try:
+            long_wf_id = match_short_id(wf_id)
+            long_wf_ids.append(long_wf_id)
+        except ClientError as err:
+            logging.info(f"Could not match workflow ID {wf_id}: {err}")
+
+    return long_wf_ids
 
 
 def get_wf_status(wf_id):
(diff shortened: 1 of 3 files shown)


beeflow/client/bee_client.py
```diff
@@ -307,6 +314,39 @@ def match_short_id(wf_id):
...
+            i += 1
+        if len(matches) == 1:
+            results.append(matches[0])
+        elif len(matches) > 1:
+            print(
+                f"User provided workflow id {short_id} is ambiguous "
+                "and matches multiple stored workflow IDs:"
+            )
+            for m in matches:
+                print(f" - {m}")
```
jeffreyt, line 342: Just want to confirm, let's say: 
- I pass a list of `wf_ids` I want to modify.  
- Some have only one match, and some have multiple matches.  
- The operation will continue for ids with one match but anything with ambiguous multiple matches will be excluded.  

Is this correct and if so do we have a test to cover this case?

kabir-vats: This is correct, added a test case

pagrubel (changes requested):
This looks really good. Would you please add something to the documentation for the option of cancelling and removing multiple workflows. 
It is very useful. I was even able to remove all workflows except the one that was running.
The feature of listing all the workflows with the ambiguous id is nice. 
All, well thought out, with nice info when used incorrectly.

You might fix this misspelling as you add the documentation:
"Workflow 6c11a3 removeed!"

docs/sphinx/commands.rst
```diff
@@ -82,22 +82,22 @@ Arguments:
 ``beeflow pause``: Pause a workflow (Running tasks will finish)
```
pagrubel:
```suggestion
``beeflow pause``: Pause a workflow(s) (Running tasks will finish)
```

pagrubel (approved):
This looks good. I also ran multiple tests. There is a problem when cancelling workflows but that existed before and is in another issue. All the options work well and thank you for streamlining methods.

pagrubel (approved):
This looks good and runs well. Thank you for streamlining the requests.

PR #1147, opened 2025-11-04
Issue #1045 beeflow list output columns lineup

### LINKED ISSUE #1045: If name is too long then beeflow list doesn't line up
Whitespace is misaligned if name is too long. I assume its just using a simple ``\t``, eg.:
```
Name	ID	Status
myworkfl	0d3d5f	Archived
```

jeffreyt:
Fixes #1045
- Change list to use tabulate instead of \t tabs

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -81,6 +81,7 @@ Contributors:
 * Jacob Tronge - `jtronge <https://github.com/jtronge>`_
 * Kabir Vats - `kabir-vats <https://github.com/kabir-vats>`_
 * Sare Beste Zeytun - `sbzeytun <https://github.com/sbzeytun>`_
+* Jeff Tamashiro - `jeffreyt <https://github.com/jeffreyt>`_
 
 Concept and Design Contributors
 
--- a/beeflow/client/bee_client.py
+++ b/beeflow/client/bee_client.py
@@ -612,8 +612,8 @@ def unpackage(package_path, dest_path):
 def list_workflows():
     """List all workflows."""
     workflow_list = get_wf_list()
-    headers = [typer.style(h, fg=typer.colors.GREEN) for h in ["Name", "ID", "Status"]]
     if workflow_list:
+        headers = [typer.style(h, fg=typer.colors.GREEN) for h in ["Name", "ID", "Status"]]
         data = []
         for wf_info in workflow_list:
             data.append([wf_info.wf_name, _short_id(wf_info.wf_id), wf_info.wf_status])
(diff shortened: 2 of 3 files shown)


wmason-lanl:
As is the change looks good to me.  

A few things to consider but not essential for fixing the issue:
- Do we want to support disabling the header color with the `color=False` flag to typer.style if [NO_COLOR](https://no-color.org/) is set?
  - Should the header be tabulated before adding color?  Right now the header color is added and then the header/data is tabulated.  If we want to support disabling header color, the ANSI escape codes will be counted as additional characters which would break alignment.  If color is added to the header after the table generation, it can be selectively enabled and won't break alignment.
- When you take into consideration the most recent commit to develop ([8cae0b](https://github.com/lanl/BEE/commit/8cae0b5b8ad3ace2d10403c33eb6fe840699ec6c)) to "use user provided wf name instead of cwl file name" what happens if you have a very long workflow name or the name contains a new line?
  - Should a column width limit be set?  `tabulate` allows for `maxcolwidths`.
  - What about newline or other control characters?  Should the workflow name be stripped or otherwise sanitized?

pagrubel:
@jeffreyt Please add yourself as a contributor in the README.rst

jeffreyt:
> As is the change looks good to me.
> 
> A few things to consider but not essential for fixing the issue:
> 
>     * Do we want to support disabling the header color with the `color=False` flag to typer.style if [NO_COLOR](https://no-color.org/) is set?
>       
>       * Should the header be tabulated before adding color?  Right now the header color is added and then the header/data is tabulated.  If we want to support disabling header color, the ANSI escape codes will be counted as additional characters which would break alignment.  If color is added to the header after the table generation, it can be selectively enabled and won't break alignment.
> 
>     * When you take into consideration the most recent commit to develop ([8cae0b](https://github.com/lanl/BEE/commit/8cae0b5b8ad3ace2d10403c33eb6fe840699ec6c)) to "use user provided wf name instead of cwl file name" what happens if you have a very long workflow name or the name contains a new line?
>       
>       * Should a column width limit be set?  `tabulate` allows for `maxcolwidths`.
>       * What about newline or other control characters?  Should the workflow name be stripped or otherwise sanitized?

These are really good valid inputs but probably outside the scope of this simple fix.  The sanitation and character limiting is something that probably deserves its own issue since it should happen when the user initially inputs data and could work globally across every input.

wmason-lanl (approved):
Additional changes look good.

PR #1148, opened 2025-11-06
1092 posixpath task workdirs

### LINKED ISSUE #1092: Task workdirs cannot be a PosixPath
Specifying a workdir for a `Task` using `TaskReq(workdir=workdir)` only works if `workdir` is a string, you cannot use a `PosixPath` object.

wmason-lanl:
Fixes #1092 

--
Task workdirs cannot be a PosixPath (#1092)
    
* Accept workdir as a string or a PosixPath
* Parametrize test_workflow_task_req pytest for string or PosixPath workdir

### DIFF
--- a/beeflow/common/cwl/cwl.py
+++ b/beeflow/common/cwl/cwl.py
@@ -1,7 +1,7 @@
 """Create and manage CWL files."""
 from dataclasses import dataclass
 from io import StringIO
-from pathlib import Path, PosixPath
+from pathlib import Path
 from typing import Optional, Union
 import ruamel.yaml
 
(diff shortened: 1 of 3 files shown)


beeflow/common/cwl/cwl.py
```diff
@@ -457,7 +458,13 @@ def __repr__(self):
 class TaskRequirement:
     """Defines task requirement."""
 
-    workdir: str
+    # Accept either a string or a PosixPath
+    workdir: Union[str, PosixPath] = None
+
+    def __post_init__(self) -> None:
+        """ Convert a Path object to a string."""
+        if isinstance(self.workdir, (Path, PosixPath)):
```
jeffreyt: You have Path here alongside here so is there any reason to single out `PosixPath`?  Couldn't we just keep the same logic and just use Path generically which would then accept all the subclasses including `PosixPath`.  For example if you did:
```
p = PosixPath('.')
isinstance(p, Path) # would still return true
```

wmason-lanl: I was just focused on the request in the issue which asked for `PosixPath`.  Updated to use general `Path` instead of the specific `PosixPath`.  Does the new [commit](https://github.com/lanl/BEE/pull/1148/commits/e866bb6d4bf7161b5c38a8142abf0227fd68902a) look better?

jeffreyt: Yeah, looking at the issue I can see why you did it that way.  But your changes look really good and I think covers the issue as well as gives some flexibility for that path input.

pagrubel:
@wmason-lanl This looks good, please merge develop and fix any conflicts.

pagrubel (approved):
Looks good, thank you!

PR #1150, opened 2025-11-13
use user provided wf name instead of cwl file name

### LINKED ISSUE #1149: Use User-Provided Workflow Name Instead of CWL Filename
BEE currently uses the filename of the cwl file (for example workflow.cwl or main.cwl) to set the wf_name parameter. This is not very intuitive, and doesn't give users the ability to dynamically set workflow names when submitting jobs.

kabir-vats:
Resolves #1149


pagrubel (approved):
This looks good thank you for cleaning up the client.

Issue #1152, opened 2025-11-25
Checkpoint/Restart when job does not fail but still needs a restart


pagrubel:
In some cases the user looks for an alert and cleanly ends the job but wants a restart depending on the presence or absence of a sentinel file.
So restarting the job needs:
- [ ] To continue submitting jobs until a sentinel file specified is found or not, as specified by something like adding the following to the requirement
        restart_on_file_exists: True - restart if file exists, False - restart if file does not exist

- [ ] In some cases only restart if job fails and use the above condition

- [x] Also, remove container_path (but check that it is not used) done in PR #1153

pagrubel:
Right now Checkpoint/Restart only works when a job fails.

rstyd:
Current version

```
beeflow:CheckpointRequirement:
     enabled: true
     file_path: checkpoint_output_dir
     file_regex: backup[0-9]*.crx
     restart_parameters: -R
     num_tries: 3
```

New Version

```
beeflow:CheckpointRequirement:
     enabled: true
     checkpoint_output_dir: checkpoint_output_dir
     sentinel_file_path: sentinel_file_path
     # Restart if the sentinel file does exist
     restart_on_file_exists: True
     # Restart if the sentinel file does not exist
     restart_on_file_exists: False
     # Only restart when a slurm job fails or timesout 
     restart_on_failure: True (Default true)
     file_regex: backup[0-9]*.crx
     restart_parameters: -R
     num_tries: 3
```

rstyd:
Most of the code changes for this will just be in [background.py](http://background.py/) within the task_manager module in the update_jobs function when it handles state changes

https://github.com/lanl/BEE/blob/c721ce0b8cf03544f72be6f1f3db0884708d48f0/beeflow/task_manager/background.py#L105-L124

For the restart_on_failure parameter, we’ll need to modify https://github.com/lanl/BEE/blob/c721ce0b8cf03544f72be6f1f3db0884708d48f0/beeflow/task_manager/background.py#L108 

We’ll need to change the new_job_state check to use the COMPLETED_STATES set defined at the top of the file so it gets called for all completed states and not just failed or timeout jobs. https://github.com/lanl/BEE/blob/c721ce0b8cf03544f72be6f1f3db0884708d48f0/beeflow/task_manager/background.py#L29 

[...]

PR #1153, opened 2025-11-26
Remove container_path from Checkpoint Requirements, not used

pagrubel:
The checkpoint requirement does not need a container path and it is not used, so deleting it. This is the first part of fixing issue #1152


rstyd (approved):
Approved!

PR #1157, opened 2025-12-09
Support spack installed redis-server instead of containerized version

jeffreyt:


beeflow/client/core.py
```diff
@@ -254,13 +254,6 @@ def celery():
     # Run this before daemonizing in order to avoid slow background start
     # container_path = paths.redis_container()
     # If it exists, we assume that it actually has a valid container
-    # if not os.path.exists(container_path):
-        # print('Unpacking Redis image...')
-        # subprocess.check_call(['ch-convert', '-i', 'tar', '-o', 'dir',
-        #                       bc.get('DEFAULT', 'redis_image'), container_path])
-    if not container_manager.check_container_dir('redis'):
```
pagrubel, line 261: Where are you checking if I want to use the redis container?

beeflow/common/config_driver.py
```diff
@@ -277,6 +281,10 @@ def unique_port():
                  default=REDIS_IMAGE, info='redis container image',
                  input_fn=filepath_completion_input, prompt=True)
 
+VALIDATOR.option('DEFAULT', 'spack_path', validator=validation.dir_,
+                 info='Spack environment path',
```
pagrubel:
```suggestion
                default='.', info='Spack environment path',
```
This prevents an error when the user just hits return, None does not work.
This would also help the user that is not using spack.

pagrubel (commented):
I hit an error when running beecofig new. 
Also, before this beeconfig new did not need query for any answers. I think it would be best if you add the spack_path using the interactive option or --spack_path.

pagrubel (approved):
This looks good and works well for both cases of using redis container or a spack installed version.

PR #1160, opened 2025-12-15
1152 Checkpoint/Restart when job does not fail but still needs a restart

### LINKED ISSUE #1152: Checkpoint/Restart when job does not fail but still needs a restart

wmason-lanl:
Fixes [#1152 ](https://github.com/lanl/BEE/issues/1152)

--
Checkpoint/Restart when job does not fail but still needs a restart ([#1152](https://github.com/lanl/BEE/issues/1152))

- Initial commit.


pagrubel:
This looks great. I tried out the integration tests manually and they worked fine. At some point they should be added to github actions since they are relatively short. Thank you for your hard work.

pagrubel (approved):

Issue #1165, opened 2026-01-06
If slurmrestd is not use or available on a system an error occures in validation


pagrubel:
If on a system where slurmrestd is unvailable an error occurs. 
To duplicate this on a system edit the bee.conf to set  use_commands = True

The current work around is to delete these lines in bee.conf when using commands instead of slurmrestd
```
[slurm attributes]
attributes = partition,nodes
```
They should probably be ignored if using commands, so the error won't occur.

The error is: 
```
beeflow core start
Traceback (most recent call last):

  File "/vast/home/pagrubel/beedev-env/bin/beeflow", line 6, in <module>
    sys.exit(main())
             ^^^^^^

  File "/vast/home/pagrubel/BEE/BEE/beeflow/client/bee_client.py", line 994, in main
    app()

  File "/vast/home/pagrubel/beedev-env/lib/python3.12/site-packages/typer/main.py", line 289, in __call__

  File "/vast/home/pagrubel/beedev-env/lib/python3.12/site-packages/typer/main.py", line 280, in __call__

  File "/vast/home/pagrubel/beedev-env/lib/python3.12/site-packages/click/core.py", line 1157, in __call__

  File "/vast/home/pagrubel/beedev-env/lib/python3.12/site-packages/click/core.py", line 1078, in main

  File "/vast/home/pagrubel/beedev-env/lib/python3.12/site-packages/click/core.py", line 1688, in invoke

  File "/vast/home/pagrubel/beedev-env/lib/python3.12/site-packages/click/core.py", line 1688, in invoke

  File "/vast/home/pagrubel/beedev-env/lib/python3.12/site-packages/click/core.py", line 1434, in invoke

  File "/vast/home/pagrubel/beedev-env/lib/python3.12/site-packages/click/core.py", line 783, in invoke

  File "/vast/home/pagrubel/beedev-env/lib/python3.12/site-packages/typer/main.py", line 607, in wrapper

  File "/vast/home/pagrubel/BEE/BEE/beeflow/client/core.py", line 451, in start
    if bee_client.get_hostname() == "" and bee_client.check_backend_status() == "":
       ^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/vast/home/pagrubel/BEE/BEE/beeflow/client/bee_client.py", line 94, in get_hostname
    db = bdb.connect_db(client_db, db_path())
                                   ^^^^^^^^^

[...]

PR #1166, opened 2026-01-06
Fix validation error when switching from slurmrestd in config

sbzeytun:
This is a temporary fix for issue #1165. Although this resolves the error message, there are still issues regarding the config file when switching between schedulers. For instance with the "beeflow config new -a (--attributes)" command, it defaults back to the slurmrestd attributes even if we switch from slurmrestd to command line. Additionally, the "[slurm attributes]" section is not deleted, rather it is ignored when switching to the command line and using "[slurm command attributes]" (Not a big issue, however it may be confusing). After switching over, it does use the slurm command line attributes.


sbzeytun:
This issue was addressed in PR #1167.

PR #1167, opened 2026-01-07
Fix Slurm attribute errors when switching between schedulers and choosing new attributes

sbzeytun:
This is for issue 1165. This should fix the validation error and ensure the correct attributes section shows up when you change use_commands from False to True, and vice versa. The previously used attribute section should be deleted to ensure there's no confusion. This should also fix the issue when using "beeflow config new -a." Previously, it defaulted to show slurm attributes even when using the command-line interface. Make sure after editing the config file to stop and start beeflow to see the changes.


pagrubel (approved):
This looks great. Thank you for fixing. We are now able to do add some requested items that are in the attributes.

PR #1172, opened 2026-01-27
Fix Slurmrestd Version Inference (#1168)

### LINKED ISSUE #1168: Fix Slurmrestd Version Inference
The function `get_slurmrestd_version` needs to be upated
https://github.com/lanl/BEE/blob/83fe32ac532e549984e7b656527f20c128f5763e/beeflow/common/worker/utils.py#L41-L54

The issue is that the way I was getting the api_version through `slurmrestd -s list` no longer works since slurm 24.11.x no longer uses versioned apis. So, instead we need to get it from slurmrestd -d list which looks like 
```
Possible data_parser plugins:
data_parser/v0.0.39
data_parser/v0.0.40
data_parser/v0.0.41
``` 

Instead of `slurmestd -s list` output which looks like 

```
Possible OpenAPI plugins:
openapi/dbv0.0.39
openapi/slurmctld
openapi/slurmdbd
openapi/v0.0.39
``` 

So it's really just a matter of updating the subprocess call and changing the regex, and that _should_ do it.

wmason-lanl:
Fixes [#1168](https://github.com/lanl/BEE/issues/1168) 

--
Fix Slurmrestd Version Inference (#1168)

* Updated get_slurmrestd_version function to query api_version with new syntax

* Updated job status query in SlurmrestdWorker

* Updated test_slurm_worker.py with new slurmrestd syntax

### DIFF
--- a/.github/workflows/testing-coverage.yml
+++ b/.github/workflows/testing-coverage.yml
@@ -23,7 +23,7 @@ jobs:
       BEE_WORKER: ${{ matrix.bee_worker }}
     # Note: Needs to run on 22.04 or later since slurmrestd doesn't seem to be
     # available on 20.04
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v4
       - name: Install and Configure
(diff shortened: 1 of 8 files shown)


pagrubel (approved):
The tests fail due to the use Ubuntu for our github actions, even the latest version of Ubuntu does not have a recent enough version of slurmrestd. These tests work in our internal testing. 

An alternated to slurmrestd would be to use slurm commands, still supported.

PR #1176, opened 2026-02-03
Add more data artifacts when CI fails

jeffreyt:
- ci/env.sh sets a bunch of useful variables, after the first time it runs, capture them in $GITHUB_ENV which makes them accessible downstream
- downstream add a new step in case of failure to dump them in a single directory, modify following step to save the single directory as an artifact 

- So far I added:
  - individual process logs (wfm, redis...)
  - bee config file
  - slurm config/logs
  - db files

We can also add anything else we may need to examine here.

### DIFF
--- a/.github/workflows/testing-coverage.yml
+++ b/.github/workflows/testing-coverage.yml
@@ -51,12 +51,12 @@ jobs:
         if: failure()   # only when the job fails
         run: |
           mkdir bee-error-logs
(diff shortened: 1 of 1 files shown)


jeffreyt:
rebased with develop

pagrubel (approved):
Thanks, I was able to test this with failing tests of another PR.

PR #1178, opened 2026-02-04
Fix issue when slurmrestd is not in environment

rstyd:
This PR fixes the issue that has recently occurred when slurmrestd does not exist in a user's path.


pagrubel:
Can you make a PR that doesn't include changing the redis management code, so it is cleaner to review?

pagrubel:
Please start the branch from a fresh checkout of develop.

PR #1182, opened 2026-03-10
Change default attributes

pagrubel:
This PR adds command and job id to defaults slurm attributes  and id to the flux defaults.
Note: there is no current command attribute for flux


rstyd (approved):
Approved!

PR #1206, opened 2026-04-17
Enable BEE to Run a Workflow Directly from a Provided Sbatch

rstyd:
This PR implements running workflows directly from an sbatch script. The one current requirement is that the sbatch script needs to include a ${error} and ${output} in the directives:

```
#SBATCH --output=${output}
#SBATCH --error=${error}
``` 

This is so BEE can save the stderr and stdout files to the appropriate location.


pagrubel:
Why are checks not running on this?

pagrubel (dismissed):
Please add documentation of the new requirement to docs/sphinx/bee_cwl.rxt
I'll continue to review it. run.sh didn't work without modifications.

docs/sphinx/bee_cwl.rst
```diff
@@ -349,6 +353,20 @@ Contensts of slurm_conf.json::
      "reservation": "reservation-a"
     }
 
+Alternate:: 
+
+    beeflow:Slurm Requirement:
+        sbatch: "run_sbatch.sh"
+
+
+Contents of run_sbatch.sh::
```
pagrubel, line 362:
```suggestion
Contents of run_sbatch.sh::

```

pagrubel (dismissed):

pagrubel:
Please fix unit tests

beeflow/common/cwl/examples/comd_sbatch.py
```diff
@@ -0,0 +1,38 @@
...
+import pathlib
+from beeflow.common.cwl.workflow import (Task, Input, Output, MPI, Charliecloud,
+                                         Workflow, Slurm, Script)
+
+
+def main():
+    """Recreate the COMD workflow."""
+    # Specifies the comd task
+    comd_task = Task(name="comd",
+                     base_command="/CoMD/bin/CoMD-mpi -e",
```
jeffreyt: I tried this with some simple examples and it worked well.  I'm just wondering, do you have to put something for the `base_command`?  Is this affecting the run in any way?

rstyd: Hi Jeff, 

That's a great question! We should probably make it optional if the user specifies sbatch.

pagrubel: So are you saying it is not optional now?  Is it ignored?

rstyd: Right now it just gets set, but doesn't actually do anything. I'm pushing a change so it just sets the base_command to a placeholder which is just the command "true". This still has some awkwardness with our current CWL specification approach since the stdout and stderr for example are also going to be ignored, but those don't need to be specified by default.

pagrubel:
I wrote a cwl generator to test what happens when a script doesn't exist but a step in the workflow writes one to be used in a task later. The CWL parser looked for the script and could not find and gave an error. I was able to get it running by having a blank file of that name available. Then all worked well.

beeflow/common/worker/utils.py
```diff
@@ -85,3 +86,43 @@ def format_start_time(start_time):
...
+                stderr = token.split("=", 1)[1]
+            elif token in ("--error", "-e"):
+                stderr = next(it, None)
+    if stdout is None and stderr is None:
+        stdout = default_output
+        stderr = default_output
+    elif stdout is not None and stderr is None:
+        stderr = stdout
+    elif stdout is None and stderr is not None:
+        stdout = default_output
```
pagrubel:
```suggestion
    if stdout is None:
        stdout = default_output
    if stderr is None:
        stderr = stdout
```

beeflow/common/cwl/examples/comd_sbatch.py
pagrubel, line 1: Should this even be an example? It just uses a batch file that makes no sense.

beeflow/common/cwl/examples/run.sh
pagrubel, line 1: This file won't run as is and unless there is some information about it, I think it should be deleted

beeflow/common/cwl/examples/write_script.py
pagrubel, line 1: If inputs and base_command are not needed they should be deleted from the second step.

pagrubel (commented):
This looks good. I just had some minor recommendations.

beeflow/common/worker/utils.py
pagrubel: Less confusion, just need to clarify that stdout is decided first.

beeflow/common/cwl/examples/comd_sbatch.py
pagrubel: I deleted it.

pagrubel (approved):
This looks good and works well. I was able to run a workflow that writes a script in one step and run it in another, later step. The only caveat that we need to consider for the future is if we don't know the name of the script a priori, that can pose problems.

Issue #1141, opened 2025-10-17
Enable beeflow core reset when beeflow is not running


rstyd:
`beeflow core reset` is intended to stop all the BEE components and delete the bee_workdir directory. Currently, this only works if `beeflow` has is running, but there are plenty of use cases where beeflow has been stopped but we still want to reset. This code should be changed so that it'll support doing a reset when beeflow is not running.

PR #1210, opened 2026-04-28
Enable beeflow core reset when beeflow is not running (#1141)

wmason-lanl:
* Check if beeflow socket exists and is responsive

* Get the workflow list if beeflow is actually running

* Conditional stop; proceed with removing directory.


pagrubel:
This works well I would like to suggest we add the information that they can do an archive I'll add it to the file as a suggestion.

beeflow/client/core.py
```diff
@@ -616,9 +633,13 @@ def reset(archive: bool = typer.Option(False, '--archive', '-a',
     warn(f"\n    A reset will remove this directory: {dir_to_delete}\n")
     if archive:
         print("    Archive flag is set: logs, workflows and containers will be backed up.")
+
```
pagrubel, line 639:
```suggestion
else:
    warning = "If you want to save logs, workflows and containers,"
    warning += " answer no and rerun with --archive flag."
    warn(warning)

```

pagrubel:
This looks good. I plan to merge it as soon as the tests pass. I did add text to inform the user of the --archive option.

pagrubel (approved):

PR #1212, opened 2026-04-30
Add documentation builds for tags

pagrubel:
This PR should add documentation for tags and when we push a release to main.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -4,16 +4,76 @@ on:
   workflow_dispatch: {}
   push:
     branches: [main, develop]
+    tags:
+      - '*'
   pull_request:
     types: [opened, synchronize, edited, labeled, unlabeled]
     branches: [main, develop]
+  delete:
+
+permissions:
+  contents: write
 
 jobs:
   docs:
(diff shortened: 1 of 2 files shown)


.github/workflows/docs.yml
```diff
@@ -4,16 +4,76 @@ on:
...
     types: [opened, synchronize, edited, labeled, unlabeled]
     branches: [main, develop]
+  delete:
+
+permissions:
+  contents: write
 
 jobs:
   docs:
-    if: ${{ !(contains(github.event.pull_request.labels.*.name, 'WIP (no-ci)')) && !(contains(github.event.pull_request.labels.*.name, 'WIP (lint-only)')) }}
```
jeffreyt, line 13: So does this mean we will run docs even with WIP label now?

pagrubel: I suppose I could put that back, but it doesn't publish them anyway

pagrubel:
Yeah I changed the label and it runs the docs. I'll put it back

.github/workflows/docs.yml
jeffreyt: Ok, if it doesn't publish that's probably ok.

pagrubel: It won't run the doc for WIP (lint-only) or (no-ci) now

jeffreyt (approved):

jeffreyt (approved):

PR #1218, opened 2026-05-14
Allow baremetal jobs

kchilleri:
This PR adds a new beeflow requirement that allows the user to specify per-task execution behavior. An individual task can now run locally (through the Simple worker), or be submitted through a workload scheduler such as Slurm or Flux.

This requirement is optional. If it is not provided, `beeflow` uses the configured default workload scheduler, preserving the usual behavior.

Implementation Approach:
-  Added CWL extension for execution mode
    - Created beeflow:WorkloadRequirement hint/requirement in CWL parser
    - Supports mode: baremetal for local execution or mode: scheduler for HPC schedulers    
- Dynamic worker selection in Task Manager
    - Implemented scheduler_for_task() to inspect task requirements and select appropriate worker (Simple, Slurm, Flux)
    - Extended database schema to track which scheduler each job uses
- Reused SimpleWorker for baremetal execution
- Process group management for robust cleanup
    - Used start_new_session=True in subprocess.Popen() to make spawned process a session leader
    - Enabled os.killpg(pgid, signal.SIGTERM) to kill entire process group
    - Ensures child/forked processes are properly terminated when parent task is cancelled
- File-based status tracking
    - Implemented status directory structure: ~/.beeflow/workflows/{workflow_id}/simple_worker_status/{pid}.returncode
    - Script writes return code to file upon completion. This preserves the return code and also avoids race conditions. 
- Return code validation
    - Distinguishes COMPLETED (rc=0), FAILED (rc≠0), and RUNNING (no returncode file)


kchilleri:
@rstyd @pagrubel this PR should be ready to review. The last thing I want to do is replicate Chris' data download example.

docs/sphinx/bee_cwl.rst
```diff
@@ -304,6 +305,44 @@ Example pre-script to preserve outputs during Checkpoint/Restarts (bash)
...
+
+**Example 2: Force specific scheduler**
+
+Override default scheduler for a specific task::
+
+    hints:
+      beeflow:WorkloadRequirement:
+        mode: scheduler
+        scheduler: Flux
+
```
pagrubel, line 349: Maybe add a third example for a task on baremetal like example 2.

beeflow/task_manager/utils.py
```diff
@@ -35,6 +35,47 @@ def connect_db():
...
+        if mode == 'baremetal':
+            selected_scheduler = 'Simple'
+        elif mode == 'scheduler':
+            selected_scheduler = wl_requirement.get('scheduler', default_sched)
+            if not isinstance(selected_scheduler, str):
+                raise RuntimeError(
+                    'beeflow:WorkloadRequirement scheduler must be a string.'
+                )
+        else:
+            raise RuntimeError(
```
pagrubel, line 65: This works, but is not indicated to the user, the state of the task just remains in the SUBMIT state. Is there some way that the error could be caught during parse?

pagrubel:
@kchilleri Using an unsupported mode for WorkloadRequirement produces an uncaught exception and a full traceback. Could the mode be validated when the CWL is parsed, or could this exception be caught and reported as a concise user-facing error? It fails before submission.  It would also be good to add a test for an invalid mode.

pagrubel:
Also the metadata needs improvement but that can be another PR. Right now it refers to our temporary workflows directory.

kchilleri:
> @kchilleri Using an unsupported mode for WorkloadRequirement produces an uncaught exception and a full traceback. Could the mode be validated when the CWL is parsed, or could this exception be caught and reported as a concise user-facing error? It fails before submission. It would also be good to add a test for an invalid mode.

@pagrubel @rstyd This PR is ready for review again. I have removed the full traceback returned to the terminal with an invalid WorkloadRequirement mode/scheduler during parse time. Now we print a user-friendly error message. Additionally, there is already a test that checks for invalid argument values.

pagrubel (approved):
This is great! I tested several scenarios and the failures are now easy to figure out from a user perspective. Thanks for all the hard work on this.

rstyd (approved):
Looks great to me

Issue #1192, opened 2026-04-07
Query does not update current metadata


pagrubel:
Apparently, BEE does not update the metadata upon polling the status of a task until the task state changes. This does not give the proper runtime information for a user query while the job is running.
This example is longer running than in the examples and using slurm commands not slurmrestd (doesn't have runtime)
```
beeflow query 522
Running
task_name    task_state       JobId  Partition    RunTime    NodeList    Command
-----------  ------------  --------  -----------  ---------  ----------  ---------------------------------------------------------------
clamr        RUNNING       16953860  general      00:00:04   cn243       /...l/beeworkdir/workdir/clamr-09ae/clamr-09ae.sh
ffmpeg       WAITING

$ squeue -u p...
             JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
          16953860   general clamr-09     p...      R       1:28      1 cn243

$ beeflow query 522
Running
task_name    task_state       JobId  Partition    RunTime    NodeList    Command
-----------  ------------  --------  -----------  ---------  ----------  ---------------------------------------------------------------
clamr        COMPLETING    16953860  general      00:01:47   cn243       /..../beeworkdir/workdir/clamr-09ae/clamr-09ae.sh
ffmpeg       WAITING

$ squeue -u pagrubel
             JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
          16953860   general clamr-09     p...     CG       1:47      1 cn243
```

sbzeytun:
This issue was satisfied by PR #1222

Issue #1133, opened 2025-08-26
Change Format of Improved Status


rstyd:
Need to rework Beste's improved status work to use YAML for the formatting.

pagrubel:
Also there are additional consideration in issue #1130. This is one of them. I like the idea of making small PR's though, just make sure we check this off in that one when complete.

sbzeytun:
This was satisfied by PR #1222

PR #1222, opened 2026-06-08
Update metadata correctly when querying and save metadata as yaml #1192

sbzeytun:
Resolves Issues #1223,  #1192,  #1133

Ensure metadata gets properly updated so that users don't see stale information, such as runtime. For consistency, I also changed the saved format so that the metadata for each task gets saved as a yaml file. This also removes the unnecessary information displayed in the top and bottom lines of the metadata yaml file.


pagrubel (approved):
This works well esp. for the slurm commands case and the default queries. Thank you.

pagrubel:
This works well esp. for the slurm commands case and the default queries. Thank you.

Issue #1225, opened 2026-06-12
Refactor slurmrestd attributes when querying


sbzeytun:
I noticed there were some weird things occurring with certain attributes when you query with slurmrestd. Many important attributes were not showing up when the user sets them in the config. When the attributes do not show up, the formatting was off and the specific attribute information seems to shift visually when querying. This seems to only be a problem for slurmrestd, but I am looking out for any related problems with the other schedulers. 

<img width="2203" height="280" alt="Image" src="https://github.com/user-attachments/assets/97bae7c4-43f0-4226-9072-4b3609165cf1" />

This example occurred when I set these attributes in the config: allocating_node,cluster,start_time,command,job_state,accrue_time (exact order). Note: task_name and task_state always appear regardless of what the user sets.

sbzeytun:
This issue was satisfied with pr #1226

PR #1226, opened 2026-06-15
Fix slurm restd attr

sbzeytun:
Fixes issue #1225. I also made the time specific attributes to be consistent among the slurmrestd and slurm command line schedulers, which was something I was wanting to do last time.


pagrubel (changes requested):
It's nice to see the current running time when using slurmrestd.
However, the runtime resets to zero if the jobs is cancelled or fails or possibly times out. 

I was able to test it for cancelling and failed by running the clamr example and cancelling the job with the scancel command. Also, tested the failed state by deleting the graphics_output directory while ffmpeg was running. I haven't tested timeout but could come up with a test for that. Here are the results:

The query showed ffmpeg at 9 seconds before it failed, show should show that.


```
Archived/Partial-Fail
task_name    task_state      job_id  partition    runtime    nodes    command
-----------  ------------  --------  -----------  ---------  -------  --------------------------------------
----------------------------------------------
clamr        COMPLETED     17188568  scaling	  00:01:54   cn336    /vast/home/pagrubel/beeworkdir/clamr-f
fmpeg-build/workdir/clamr-86ef/clamr-86ef.sh
ffmpeg       FAILED 	   17188569  scaling	  00:00:00   cn366    /vast/home/pagrubel/beeworkdir/clamr-f
fmpeg-build/workdir/ffmpeg-bd91/ffmpeg-bd91.sh
```

Similarly, for this example the time was over a minute for clamr before I cancelled it. 
Think about a job that runs for hours and somehow is cancelled or fails. The user may need that information.

 beeflow query 76
```
Archived/Failed
task_name    task_state      job_id  partition    runtime    nodes    command
-----------  ------------  --------  -----------  ---------  -------  ----------------------------------------------------------------------------------
clamr        CANCELLED     17187581  scaling      00:00:00   cn339    /vast/home/pagrubel/beeworkdir/clamr-ffmpeg-build/workdir/clamr-74cd/clamr-74cd.sh
ffmpeg       DEP_FAIL
```

pagrubel (dismissed):
I tested this with a timelimit and a longer clamr example and it also clears out all time for the TIMEOUT state. There may be other states that this will not work for such as PAUSED or SUSPENDED.

pagrubel (approved):
This looks good, I did make a change to calculate time for any state.

PR #1230, opened 2026-06-25
Optimize CI tests through github's caching mechanism

sbzeytun:


wmason-lanl:
@sbzeytun Below are a few things I noticed on this PR.  I think the first one needs to be addressed.  The others are mostly suggestions or best practices.

1. I recommend updating the cache keys to reflect what is being run/installed and not depend on just the hash for './ci/env.sh'.  If you are running both `env.sh` and `deps_build.sh`, you would want the cache key to hash both of those scripts.  If you only key off the hash for `env.sh`, you'll miss any changes to `deps_build.sh` which might invalidate the cache.  You can use something like the following to include both scripts:

```
key: ${{ runner.os }}-deps-${{ hashFiles('ci/env.sh', 'ci/deps_build.sh', 'ci/deps_install.sh') }}
```

2. You probably want to have separate workspace install locations for the different caches.  Right now, you are installing both the `deps_install.sh` and `flux_install.sh` caches to the same `${{ github.workspace }}/deps`.  This means the "flux" cache will contain the "deps" cache contents too.

3. You might also want to split out the Slurm install from `deps_build.sh` and only include the Charliecloud install in that script.  If the Slurm install is moved to a separate install script like `flux_install.sh`, then Slurm won't need to be built if the "deps" cache needs to be rebuilt for the `BEE Integration Test` job.

4. Another consideration is you might want to use a more granular approach for caching by splitting out the restore (`actions/cache/restore@5`) and save (`actions/cache/save@5`) steps.  The standard combined approach is probably sufficient, but using the granular approach can make reading through the caching logic a bit more clear.

wmason-lanl (approved):

PR #1231, opened 2026-06-25
Fixes missing task object

pagrubel:
Will not attempt to access/create the metadata file until there is a task object.


beeflow/wf_manager/resources/wf_update.py
```diff
@@ -180,6 +180,9 @@ def update_task_state(self, state_update):
         task = wfi.get_task_by_id(state_update.task_id)
         wfi.set_task_state(state_update.task_id, state_update.job_state)
 
-        self.handle_metadata(state_update, state_update.task_id, wfi, task.workdir)
-        if not self.handle_checkpoint_restart(state_update, task, wfi):
-            self.handle_state_change(state_update, task, wfi)
+        # Check if task exists, since it may not have been submitted yet
+        if task:
```
kchilleri, line 184: This makes sense to me. It'd be nice to add a test that verifies these calls are skipped if task is None / does not exist.

pagrubel: added test

rstyd (approved):
Looks good to me approved!

PR #1234, opened 2026-06-30
Remove check for sbatch file existing at parse time

rstyd:
This PR resolves the issue where sbatch files must exist at workflow parse time.


pagrubel (approved):
Thanks, this looks good I just changed the example to not create a blank file before running the workflow.

PR #1238, opened 2026-07-06
working version of querying latest workflow

santiag0-3:
Changed argument of Query() to optionally accept no workflow id from user. If no workflow id is given Query() gives latest workflow submitted.


pagrubel:
@santiag0-3 This works well. Since the user may need to be reminded which workflow they are querying, it would be helpful to also list the workflow name and id like is shown for the "beeflow list" command.
 This is what it looks like now:

```
Archived
task_name    task_state      job_id  partition    runtime    nodes    command
-----------  ------------  --------  -----------  ---------  -------  -------------
clamr        COMPLETED     17193379  scaling      00:01:53   cn343    .....clamr-69b4.sh
ffmpeg       COMPLETED     17193384  scaling      00:00:49   cn356    ...fmpeg-229a.sh
```
New version should be something like this:

```
clamr       ebc01a  Archived

task_name    task_state      job_id  partition    runtime    nodes    command
-----------  ------------  --------  -----------  ---------  -------  -------------
clamr        COMPLETED     17193379  scaling      00:01:53   cn343    .....clamr-69b4.sh
ffmpeg       COMPLETED     17193384  scaling      00:00:49   cn356    ...fmpeg-229a.sh
```

pagrubel (approved):
This looks good. I think we can consolidate some of this for both cases (querying with id or just leave it blank), but that can be another PR. If the user specifies a workflow, the query only has the workflow state, for now this is fine, since they know what they are querying.

PR #1240, opened 2026-07-16
Don't update coverage.svg until merged into develop

sbzeytun:
Creates a job summary located in the workflow run page that displays the coverage report instead of the bot pushing coverage.svg to a new pull request every time.


pagrubel:
THis is a good start. It would be nice to have the coverage reported since it is available and make sure that it is only changed on pushes to develop.
Also I will add a comment in the Release document to make sure coverage tests are complete with develop has had a change before pushing out a release and changing main.

.github/workflows/testing-coverage.yml
```diff
@@ -160,43 +166,49 @@ jobs:
...
 
-      - name: Verify Changed Files
+      - name: Upload coverage badge artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage-badge
+          path: ./coverage.svg
+
+      - name: Coverage comment
         if: github.event_name == 'pull_request'
```
pagrubel, line 191:
```suggestion
        if: github.event_name == 'pull_request'
        continue-on-error:  true
```

pagrubel: I think we want to ensure forks don't fail if they cannot write a comment. We can still check logs for the coverage, if the comment fails.

pagrubel (commented):
This looks good reviewers can look at the comments to see what the current coverage is. I'm wondering why it prints a threshold of 0 though.

pagrubel:
Testing is now failing due to a network problem

pagrubel (approved):
This looks good and will prevent pushing the coverage badge in the PR, while esuring it doesn't fall below the threshold. I tested it and it does show the coverage in a prominent place. For external PR's the comment may fail, however that does not fail the PR tests and reviewers can check the current coverage in the test output. It also fixes the network issue we were having. We may want to look at newer versions for Slurm though.

Issue #1221, opened 2026-06-05
Beeflow Config New Error


rstyd:
I ran into `TypeError: expected str, bytes or os.PathLike object, not NoneType` when trying `beeflow config new` on python 3.13 on a system that doesn't have a beeflow /usr/projects. It appears to breaking because there's a general assumption that the dependency containers will be in `/usr/projects/beeflow`.

pagrubel:
Maybe we should just get rid of those containers and have instructions for using spack. It really isn't that hard. @rstyd

pagrubel:
I also get that error when I do beeflow core pull-deps, so it seems rather circular. The containers don't exist but it won't let me pull them.

PR #1241, opened 2026-07-17
#1221-fix-beeflow-config-new-error

santiag0-3:
Updating defaults for graph data base to be SQLite3 instead of running neo4j container, and checking for Spack environment instead of running Redis container.


santiag0-3:
working on Darwin so I updated neo4j image path to test if logic works if no path exits. **restore later**
also delete "(test}" in next commit

PR #1247, opened 2026-07-23
Build docs after badge gets updated in gh-pages

sbzeytun:
This fix makes it so that docs.yml is dependent on testing-coverage.yml to ensure that badge gets updated in gh-pages before building the documentation. However, I made sure it can also be run on manual triggers, version releases (when tags are pushed), and on delete events.

### DIFF
--- a/.github/workflows/testing-coverage.yml
+++ b/.github/workflows/testing-coverage.yml
@@ -241,12 +241,12 @@ jobs:
       - name: Install BEE and Build Docs
         run: ./ci/docs.sh
 
-     - name: Publish latest docs to root
-       if: github.ref == 'refs/heads/main'
-       uses: peaceiris/actions-gh-pages@v4
-       with:
-         github_token: ${{ secrets.GITHUB_TOKEN }}
-         publish_branch: gh-pages
-         publish_dir: ./docs/sphinx/_build/html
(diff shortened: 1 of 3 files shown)


pagrubel:
@sbzeytun We still want to test building the docs on pull requests, sorry I may have not made that clear. We just want to do it after the coverage test whether it fails or passes. As long as we don't publish docs that would be fine.

.github/workflows/testing-coverage.yml
```diff
@@ -218,3 +218,35 @@ jobs:
           publish_dir: ./badge-publish
           destination_dir: badges/${{ github.ref_name }}
           keep_files: true
+
+  build-and-publish-docs:
+    if: ${{ always() }}
+    needs: [coverage]
+    name: Build Docs
+    runs-on: ubuntu-latest
```
pagrubel, line 226:
```suggestion
    runs-on: ubuntu-latest
    permissions:
        contents: write
```

pagrubel (commented):
Need to make sure when the documentation is built for main it can be written to gh-pages.

pagrubel (approved):
This looks good and runs the documents build. I'll open a separate PR concerning publishing for main. That only needs to be done before a release goes out.

PR #1249, opened 2026-07-27
checks for spack and sqlite3

santiag0-3:
Check for sqlite3 and Spack environment with Redis installed.


pagrubel:
From discussion at today's meeting: We should probably pin down a version of redis in the spack recipe. We may also need to instruct on using a release version of spack and some more information about spack issues in the documentation? Although I think the link helps.

PR #1250, opened 2026-07-27
Fix main docs publication

pagrubel:
Removing duplicate publication for docs when pushing to main


sbzeytun (approved):
This looks good to me

PR #1253, opened 2026-08-13
Make cache keys more descriptive and cleanup caches on closed prs

sbzeytun:
As a security layer, each opened pull request has its own cache that cannot be used by develop or other feature branches. Although unused caches are deleted after seven days, I added a new workflow that deletes caches once a pull request is closed. This reduces the noise and saves storage, since caches can pile up for each feature branch depending on what is changed (there are 17 caches right now). This will prevent having to manually delete an old, lingering cache after a pull request is closed before the 7 days.

### DIFF
--- a/.github/workflows/cleanup-cache.yml
+++ b/.github/workflows/cleanup-cache.yml
@@ -1,0 +1,29 @@
+name: Cleanup caches on closed pull requests
+on:
+  pull_request:
+    types:
+      - closed
+
+jobs:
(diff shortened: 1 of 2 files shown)


sbzeytun:
Deemed it unnecessary to include the pr reference in the cache key name since you can see it in the cache management location. Did not want to include the actual branch name since there are security issues.

pagrubel (approved):
Good idea, thank you for the PR.

Issue #1254, opened 2026-08-13
Concerning poetry environments


pagrubel:
Poetry shell command makes a virtual environment for developers and possibly causing problems with beeflow that are hard to diagnose. We have had many less problems when using poetry as installed in a python environment using pip and just using the poetry install command. We should probably get rid of the poetry information in our documentation and perhaps have a word of caution. I'm not sure the last part would be beneficial.

sbzeytun:
This was satisfied by PR #1255.

PR #1255, opened 2026-08-13
Get rid of poetry information that causes issues

sbzeytun:
This addresses issue #1254


pagrubel (approved):
This helps, thank you.

PR #1257, opened 2026-08-18
Restart Failed Workflows

rstyd:
This PR adds the ability to restart failed workflows via the retry command.

### DIFF
--- a/beeflow/client/bee_client.py
+++ b/beeflow/client/bee_client.py
@@ -38,8 +38,8 @@ from beeflow.common import paths
 from beeflow.common.object_models import generate_workflow_id
 from beeflow.client import remote_client
 from beeflow.wf_manager.models import (
-    CopyWorkflowRequest,
-    CopyWorkflowResponse,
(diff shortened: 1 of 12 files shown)


pagrubel:
@rstyd I have some more testing to do, but this works well with Failed and Cancelled tasks. However, I do have a couple of points.
Why did you remove the copy command. If it was of no use, please delete the entries for it in the documentation.
Also, since the workflow is archived even though it fails, you should at a mininum keep the previous archive and version it, to preserve provenance for all the tasks from the previous run(s). You might need to check if those are removed with the remove command too.

pagrubel (changes requested):
Should we get rid of beeflow copy? Is it used for reexecute? 
Please save the former archive with a version number or add to the original so all provenance exists.

rstyd:
Ah good point on the archive.

pagrubel:
If at all possible could you add resetting the num of tries done for checkpoint restart to zero if a task failed, giving the opportunity to do num_tries more by just retrying the workflow

Issue #1235, opened 2026-06-30
When using an the option to submit an sbatch we need to archive the script


pagrubel:
Archive sbatch script for a task that uses the SlurmRequirement sbatch.
One solution would be to copy the script into the same subdirectory the metadata goes in <task-name>-<short_id>.

pagrubel:
Also, the metadata file and the scripts should be in the archive

pagrubel:
Closed via #1259

PR #1259, opened 2026-08-19
Archive sbatch script

sbzeytun:
This address issue #1235.


pagrubel (approved):
Thanks, this works well and records the batch file as it appears when the task is run, preserving the actual run. 

I corrected the README, but it has to do with my previous error.

One thing we should probably do in the future is add a test to verify this using either use_script.py  or write_script.py (would add tests for using the cwl generator and a workflow that has a batch script) and verify the batch script in the archive as well as other expected files.

Issue #968, opened 2024-12-06
BEE needs to handle daemons on different front ends


rstyd:
Currently, if a user is on FE1 on an HPC cluster they aren't able to interact with a BEE instance on FE2. We should figure out how to accomplish this.

PR #1266, opened 2026-09-15
BEE needs to handle daemons on different front ends (#968)

wmason-lanl:
Add beeflow-ssh wrapper that automatically detects which host BEE is running on and proxies commands via SSH when needed.

- Queries client.db to determine the startup host
- Compares with current host
- If different, proxies command via SSH
- If same, runs locally with no overhead
- Handles SSH auth failures by automatically running reticket if available
- Extracts domain from current FQDN

Users can alias beeflow='beeflow-ssh' for transparent operation.


beeflow/common/scripts/beeflow-ssh
```diff
@@ -0,0 +1,158 @@
...
+
+# Get the BEEFlow working directory from configuration
+get_bee_workdir() {
+    python3 -c "from beeflow.common.config_driver import BeeConfig as bc; print(bc.get('DEFAULT', 'bee_workdir'))" 2>/dev/null
+}
+
+# Get the hostname where BEEFlow was started from the client database
+get_bee_host() {
+    local client_db="$1"
+    python3 -c "
```
davidallendj, line 22: Do we expect this to run with specific versions of Python 3?

beeflow/common/scripts/beeflow-ssh
```diff
@@ -0,0 +1,158 @@
...
+import sqlite3
+try:
+    conn = sqlite3.connect('${client_db}')
+    result = conn.execute('SELECT hostname FROM info;').fetchone()
+    if result:
+        print(result[0])
+except Exception:
+    pass
```
davidallendj, line 30: Do we want this to error silently or do we want to handle this another way?

beeflow/common/scripts/ssh_wrapper.py
davidallendj, line 1: Just curious, why use a Python wrapper instead of calling the bash directly? Is it needed to add to the `pyproject.toml`. I'm unfamiliar with how poetry works so I figured I would ask.

beeflow/common/scripts/beeflow-ssh
wmason-lanl: Yes, the BEE documentation lists Python version 3.11 to 3.14 as a requirement.

beeflow/common/scripts/beeflow-ssh
wmason-lanl: I agree it would be better to print a message that the hostname could not be found in the client database.

beeflow/common/scripts/ssh_wrapper.py
wmason-lanl: Correct.  I used the Python wrapper shim so it was easy to add to the `pyproject.toml`.

davidallendj:
After trying this out a little more, it makes me wonder if there's ever a case where you would to target specific hosts that is running BEE and only those hosts instead of all frontend nodes. Maybe adding an optional flag to allow doing something like that would be useful if that's a realistic scenario, but I'm not actually sure.

*Edit: I should probably add that if `beeflow-ssh` has its own flags that it would probably make sense to make `beeflow-ssh --help` not proxy the command and just show the relevant help message instead.*

davidallendj:
I was also wondering if the `reticket` logic is generalized enough to work with any Kerberos setup and inlined in the bash script as a function or is it site specific. If it is, then it could be tried if `command -v reticket` fails, or finally show the message to use another authentication method.

wmason-lanl:
> After trying this out a little more, it makes me wonder if there's ever a case where you would to target specific hosts that is running BEE and only those hosts instead of all frontend nodes. Maybe adding an optional flag to allow doing something like that would be useful if that's a realistic scenario, but I'm not actually sure.
> 
> _Edit: I should probably add that if `beeflow-ssh` has its own flags that it would probably make sense to make `beeflow-ssh --help` not proxy the command and just show the relevant help message instead._

I don't think you would want to target specific hosts, since only one host can be running BEE on a platform and that hostname is stored in the client database.  It servers as our source of truth.

I don't think we want `beeflow-ssh` to support its own flags, since it is a stopgap solution until we develop the functionality we are seeking.  Once we have a known functionality that we want to support, it will be absorbed into the beeflow code and the wrapper script can be removed.

wmason-lanl:
> I was also wondering if the `reticket` logic is generalized enough to work with any Kerberos setup and inlined in the bash script as a function or is it site specific. If it is, then it could be tried if `command -v reticket` fails, or finally show the message to use another authentication method.

The `reticket` script is site specific.  It isn't something generic that will work with any Kerberos setup.  In fact, there is work in progress to support the institution Kerberos TGTs.  Once that is complete, `reticket` won't be necessary.

davidallendj:
> I don't think you would want to target specific hosts, since only one host can be running BEE on a platform and that hostname is stored in the client database. It servers as our source of truth.

Right. I tried doing that right after I posted and saw that and I agree.

davidallendj:
> The `reticket` script is site specific. It isn't something generic that will work with any Kerberos setup. In fact, there is work in progress to support the institution Kerberos TGTs. Once that is complete, `reticket` won't be necessary.

I was more so thinking about generalized authentication and not too concerned about `reticket` itself. If there is a WIP for Kerberos, then I think that's the way to go.

